PR #3, opened 2018-08-30
Rework plugin broker to send broker execution result to Che-master over JSON_RPC instead of providing a REST API

garagatyi:
This PR:
- Removes usage of REST API (but REST API itself is not yet removed in the PR for the time being)
- Start Che plugins resolving process immediately after the start of the plugin broker 
- Fetches Che plugins config from a file
- Connects to the Che master by JSON_RPC over websocket
- After start/successful/unsuccessful execution sends events to JSON_RPC with the status, error, result of execution
- Add fixes to the tar processing code
- Updates dependencies of the project
- Adds more logs to understand on which step failure appeared

### DIFF
--- a/broker/broker.go
+++ b/broker/broker.go
@@ -33,21 +33,6 @@ var (
 	bus = event.NewBus()
 )
 
-// ProcessPlugins processes passed plugins metas and set success state of broker execution
-func ProcessPlugins(metas []model.PluginMeta) {
-	for _, meta := range metas {
-		err := processPlugin(meta)
-		if err != nil {
-			log.Fatal(err)
-		}
-	}
-
-	log.Println("Set success status")
-	if ok, status := storage.SetStatus(model.StatusDone); !ok {
-		log.Fatalf("Setting '%s' broker status failed. Broker has '%s' state", model.StatusDone, status)
-	}
-}
-
 // Start executes plugins metas processing and sends data to Che master
 func Start(metas []model.PluginMeta) {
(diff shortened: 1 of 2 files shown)


broker/broker.go
```diff
@@ -13,32 +13,129 @@
...
+	})
+}
+
+// PushStatuses sets given tunnel as consumer of broker events.
+func PushStatuses(tun *jsonrpc.Tunnel) {
+	bus.SubAny(&tunnelBroadcaster{tunnel: tun}, model.BrokerStatusEventType, model.BrokerResultEventType)
+}
+
+type tunnelBroadcaster struct {
+	tunnel          *jsonrpc.Tunnel
```
skabashnyuk: everything except tunnel is unused in this struct. Is this intentionally?

garagatyi:
I've aligned commit messages with the PR description and pushed with force to the PR. Sorry about that.

broker/broker.go
garagatyi: It was copied by mistake. Reconnection feature is not implemented in the broker. I've removed this code.

garagatyi:
@skabashnyuk I've removed rest and changed deps.

garagatyi:
@skabashnyuk And I've checked the broker in the walking skeleton with all those unchecked before dependencies. Everything works.

garagatyi:
@benoitf those were debug traces, I've removed them.

PR #10, opened 2018-09-27
Revert "Renamed STARTING event to STARTED and added publishing of STARTED event"

garagatyi:
Reverts eclipse/che-plugin-broker#9
PR that respects changes from #9 is not merged in Che master https://github.com/eclipse/che/pull/11344 and merging of #9 broke Che 7 flow.
And even if we merged it right now it would not help since yesterday's nightly doesn't have it.

### DIFF
--- a/broker/broker.go
+++ b/broker/broker.go
@@ -35,12 +35,11 @@ var (
 
 // Start executes plugins metas processing and sends data to Che master
 func Start(metas []model.PluginMeta) {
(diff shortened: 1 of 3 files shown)


sleshchenko:
It doesn't break anything, the only bad thing that happens is extra error logs appear. So, I wouldn't merge it since Che Server will be merged today (I hope) and tomorrow it will be needed to revert this revert =)

garagatyi:
You are right, my bad!

PR #15, opened 2018-11-19
Propagate id and version to Che Server

sleshchenko:
### What does this PR do?
Adds propagating of id and version field to Che Server.
Che Server [parses](https://github.com/eclipse/che/blob/1e78fa4d6e14f5dd0265d61cc70b26efe740d257/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/wsplugins/events/BrokerService.java#L108
) objects published by Plugin Broker as [Che Plugins](https://github.com/eclipse/che/blob/851c3a8be75b4cee278af1b98bd08c9fbb044976/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/wsplugins/model/PluginBase.java#L21-L23) which has id, version and name field. But previosly Plugin Broker didn't publish id, version and name fields and Che Server doesn't have an ability to match received configuration with Che Plugin. It may be needed like for resolving Che Plugin Commands, or containers names evaluation.

This PR doesn't add propagation of Che Plugin name because Che 7 workspace won't work since names of some of plugins doesn't suite [name format](https://github.com/eclipse/che/blob/251e5e261bac2bf4c93f113e52fd18d26b5989ab/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/wsplugins/KubernetesPluginsToolingValidator.java#L27) which Che Server requires. Example of such plugin https://github.com/eclipse/che-plugin-registry/blob/master/plugins/che-machine-exec-plugin/0.0.1/meta.yaml#L4
So, propagation of plugin name should be uncommented when https://github.com/eclipse/che/issues/11479 will be fixed.

It is related to https://github.com/eclipse/che/issues/9546

### DIFF
--- a/broker/broker.go
+++ b/broker/broker.go
@@ -170,7 +170,7 @@ func processPlugin(meta model.PluginMeta) error {
 	}
 
 	printDebug("Resolving Che plugins for '%s:%s'", meta.ID, meta.Version)
-	err = resolveToolingConfig(pluginPath)
+	err = resolveToolingConfig(&meta, pluginPath)
 	if err != nil {
 		return err
 	}
@@ -179,7 +179,7 @@ func processPlugin(meta model.PluginMeta) error {
 	return copyDependencies(pluginPath)
 }
 
-func resolveToolingConfig(workDir string) error {
+func resolveToolingConfig(meta *model.PluginMeta, workDir string) error {
 	toolingConfPath := filepath.Join(workDir, "che-plugin.yaml")
 	f, err := ioutil.ReadFile(toolingConfPath)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


skabashnyuk:
CC @l0rd

l0rd:
@sleshchenko @skabashnyuk just keep in mind that I am working on https://github.com/eclipse/che-plugin-registry/pull/63

sleshchenko:
@skabashnyuk Added tests for setStatus/addTolling/Tolling methods in Storage. I think my changes really affects only these places. Please approve if you OK with that.

PR #20, opened 2018-12-10
Change approach with multiple Dockeriles to fix Dockerhub build

garagatyi:
### What does this PR do?
Dockerhub failed to build images with previous structure even though locally everything was OK.
I changed approach to have all the dockerfiles in the root of the repo but renaming them to distinguish. 
Dockerfile for Che plugin broker now:`Dockerfile.cpb`
Dockerfile for Theia plugin broker now:`Dockerfile.tpb`
Dockerfile with CI checks now:`Dockerfile.CI`

https://github.com/redhat-developer/rh-che/issues/1134


garagatyi:
@riuvshin Please review

PR #21, opened 2018-12-11
Add init broker that removes files from /plugins

garagatyi:
### What does this PR do?
Currently, che plugin broker cleans up /plugins folder. Since the order of brokers might be affected by different reasons it brings instability. For example, Theia plugin broker adds a plugin to /plugins and then just started che plugin broker removes it.  
This PR adds init broker that is supposed to be used as init container for brokers pod.

https://github.com/eclipse/che/issues/11895


brokers/init/main.go
```diff
@@ -0,0 +1,52 @@
...
+	"log"
+	"os"
+	"path/filepath"
+
+	"github.com/eclipse/che-plugin-broker/cfg"
+	"github.com/eclipse/che-plugin-broker/common"
+	"github.com/eclipse/che-plugin-broker/model"
+)
+
+func main() {
```
sleshchenko: It would be nice to have a few words of purpose of this `init` plugin broker

brokers/che-plugin-broker/broker/broker.go
```diff
@@ -61,13 +60,6 @@ func (cheBroker *ChePluginBroker) Start(metas []model.PluginMeta) {
 	cheBroker.PubStarted()
 	cheBroker.PrintInfo("Started Plugin Broker")
 
-	// Clear any existing plugins from dir
-	log.Println("Cleaning /plugins dir")
-	err := cheBroker.ioUtil.ClearDir("/plugins")
```
amisevsk: Does it make sense to remove `ioUtil.ClearDir()`?

PR #22, opened 2018-12-20
Allow direct link to che-plagin.yaml

skabashnyuk:
### What does this PR do?
Allow the direct link to che-plagin.yaml

https://github.com/eclipse/che/issues/12246

### DIFF
--- a/brokers/che-plugin-broker/broker/broker.go
+++ b/brokers/che-plugin-broker/broker/broker.go
@@ -20,6 +20,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"gopkg.in/yaml.v2"
 
@@ -114,24 +115,34 @@ func (cheBroker *ChePluginBroker) processPlugin(meta model.PluginMeta) error {
 	if err != nil {
 		return err
 	}
+	var pluginPath string
 
-	archivePath := filepath.Join(workDir, "pluginArchive.tar.gz")
-	pluginPath := filepath.Join(workDir, "plugin")
+	if strings.HasSuffix(url, pluginFileName) {
+		chePluginYamlPath := filepath.Join(workDir, pluginFileName)
+		cheBroker.PrintDebug("Downloading plugin definition '%s' for plugin '%s:%s' to '%s'", url, meta.ID, meta.Version, chePluginYamlPath)
+		err = cheBroker.ioUtil.Download(url, chePluginYamlPath)
+		if err != nil {
+			return err
+		}
+		pluginPath = workDir
+	} else {
+		archivePath := filepath.Join(workDir, "pluginArchive.tar.gz")
+		pluginPath = filepath.Join(workDir, "plugin")
 
-	// Download an archive
-	cheBroker.PrintDebug("Downloading archive '%s' for plugin '%s:%s' to '%s'", url, meta.ID, meta.Version, archivePath)
-	err = cheBroker.ioUtil.Download(url, archivePath)
-	if err != nil {
-		return err
+		// Download an archive
+		cheBroker.PrintDebug("Downloading archive '%s' for plugin '%s:%s' to '%s'", url, meta.ID, meta.Version, archivePath)
+		err = cheBroker.ioUtil.Download(url, archivePath)
+		if err != nil {
+			return err
+		}
+
+		// Untar it
+		cheBroker.PrintDebug("Untarring archive '%s' for plugin '%s:%s' to '%s'", url, meta.ID, meta.Version, archivePath)
+		err = cheBroker.ioUtil.Untar(archivePath, pluginPath)
+		if err != nil {
+			return err
+		}
 	}
-
-	// Untar it
-	cheBroker.PrintDebug("Untarring archive '%s' for plugin '%s:%s' to '%s'", url, meta.ID, meta.Version, archivePath)
-	err = cheBroker.ioUtil.Untar(archivePath, pluginPath)
-	if err != nil {
-		return err
-	}
-
 	cheBroker.PrintDebug("Resolving '%s:%s'", meta.ID, meta.Version)
 	err = cheBroker.resolveToolingConfig(&meta, pluginPath)
 	if err != nil {
--- a/brokers/che-plugin-broker/broker/broker_test.go
+++ b/brokers/che-plugin-broker/broker/broker_test.go
@@ -21,7 +21,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
-	yaml "gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v2"
 
 	tests "github.com/eclipse/che-plugin-broker/brokers/test"
 	cmock "github.com/eclipse/che-plugin-broker/common/mocks"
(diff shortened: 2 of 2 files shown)


brokers/che-plugin-broker/broker/broker.go
```diff
@@ -114,24 +115,34 @@ func (cheBroker *ChePluginBroker) processPlugin(meta model.PluginMeta) error {
...
-	archivePath := filepath.Join(workDir, "pluginArchive.tar.gz")
-	pluginPath := filepath.Join(workDir, "plugin")
-
-	// Download an archive
-	cheBroker.PrintDebug("Downloading archive '%s' for plugin '%s:%s' to '%s'", url, meta.ID, meta.Version, archivePath)
-	err = cheBroker.ioUtil.Download(url, archivePath)
-	if err != nil {
-		return err
-	}
+	if strings.HasSuffix(url, pluginFileName) {
```
garagatyi: We have no obligation that URL should end up with a specific extension or name. And in some cases, web servers serve files, not under their real names or content can be REST API instead of a real file.
Not sure this case is so simple.

brokers/che-plugin-broker/broker/broker.go
```diff
@@ -114,24 +115,34 @@ func (cheBroker *ChePluginBroker) processPlugin(meta model.PluginMeta) error {
...
+		chePluginYamlPath := filepath.Join(workDir, pluginFileName)
+		cheBroker.PrintDebug("Downloading plugin definition '%s' for plugin '%s:%s' to '%s'", url, meta.ID, meta.Version, chePluginYamlPath)
+		err = cheBroker.ioUtil.Download(url, chePluginYamlPath)
+		if err != nil {
+			return err
+		}
+		pluginPath = workDir
+	} else {
+		archivePath := filepath.Join(workDir, "pluginArchive.tar.gz")
```
garagatyi: I tried to follow the clean code idea for the brokers. I would rather see only one level of abstraction in this function.

garagatyi: or one level of details if you will

brokers/che-plugin-broker/broker/broker.go
AndrienkoAleksandr: @garagatyi So Do you propose apply file type field to detect what we have: file or archive? As for me for quick start - suffix solution could be, than we could improve it. What Do you think?

brokers/che-plugin-broker/broker/broker.go
skabashnyuk: Ok. if I do two methods that get content. One will work with archive other with raw yaml. Will it work for you?

brokers/che-plugin-broker/broker/broker.go
skabashnyuk: @garagatyi We can change strategy and download first and then try to guess if it is an archive or not. Is that you are referring?

garagatyi: @skabashnyuk yes, I'm referring to that. But I agree with @AndrienkoAleksandr that we can start from suffix matching. Even though it is not covering all the cases it is super easy to implement.
My concerns are:
- not sure somebody would take the hard part if we have this simple suffix matching
- if unarchiving is not needed than Che master could download plugin YAML itself and not run plugin brokering

At the same time, we don't yet know how exactly brokering refactoring would be done and when, so I'm +1 to go with the proposed implementation. It doesn't harm but can benefit UX a bit.

benoitf:
besides the typo on plagin.yml I'm not so convinced. I don't think the broker should do this job. IMHO you're tring to bypass the fact that automation is not yet there.

brokers/che-plugin-broker/broker/broker.go
garagatyi: Yeah, something like:
```go
switch (getTypeOfURL()) {
  case "archive":
    processArchive()
  case "yaml":
    processYAML()
}
```
seems clean for me. WDYT?

skabashnyuk:
@benoitf 
>besides the typo on plagin.yml I'm not so convinced. I don't think the broker should do this job.

Download yaml? Download tar.gz? How strong your disagreement? 

>IMHO you're tring to bypass the fact that automation is not yet there.

Yes. I just don't like to create useless releases. What would be your suggestion here?

benoitf:
> Yes. I just don't like to create useless releases. What would be your suggestion here?

First, it's about the usecase. From what I see, your issue is about development mode so I'm not sure in development you want also to push all your changes to github as well.
I think it's kind of hijacking a process but IMHO the flow is not correct.

When you're inside a che workspace, you can use publish-script to publish to your own registry automatically and in development mode. So what's the issue you're trying to solve ?

skabashnyuk:
>First, it's about the usecase. From what I see, your issue is about development mode so I'm not sure in development you want also to push all your changes to github as well.
I think it's kind of hijacking a process but IMHO the flow is not correct.
When you're inside a che workspace, you can use publish-script to publish to your own registry automatically and in development mode. So what's the issue you're trying to solve ?

Let's review my use-case. I'm experimenting with new editor https://github.com/skabashnyuk/eclipse-che-plugins/tree/master/broadway-docker/1.0.0, I like to track all my changes, that is why I'm pushing it to git frequently.  To test this editor I'm using minishift running on OSX that is quite heavy by itself. To package editor, I need to build docker image to test each individual change, in parallel I'm developing che-editor.yaml. In this flow, I need to release che-editor.yaml or add che-editor.tar.gz each time I have a change. Yes, I can use custom registry/publish-script and deploy to it. But I don't see a reason why we should require to do that. 


>So what's the issue you're trying to solve ?

Another argument https://github.com/ws-skeleton/che-editor-theia/releases
Are they all real? Why don't we put here https://github.com/eclipse/che-plugin-registry/blob/master/plugins/org.eclipse.che.editor.theia/1.0.0/meta.yaml#L8  link to here https://raw.githubusercontent.com/ws-skeleton/che-editor-theia/master/etc/che-plugin.yaml until we developing?

brokers/che-plugin-broker/broker/broker.go
skabashnyuk: easy

skabashnyuk: @garagatyi can you take a look is this what you are asked for?

benoitf:
FYI @l0rd @slemeur 
Up to you guys

brokers/che-plugin-broker/broker/broker.go
```diff
@@ -110,11 +117,40 @@ func (cheBroker *ChePluginBroker) processPlugin(meta model.PluginMeta) error {
...
 
+	switch getTypeOfURL(url) {
+	case Archive:
+		return cheBroker.processArchive(&meta, url)
+	case Yaml:
+		return cheBroker.processYAML(&meta, url)
+	}
+	return nil
+}
+func (cheBroker *ChePluginBroker) processYAML(meta *model.PluginMeta, url string) error {
```
garagatyi:
```suggestion
 
 func (cheBroker *ChePluginBroker) processYAML(meta *model.PluginMeta, url string) error {
```

brokers/che-plugin-broker/broker/broker.go
```diff
@@ -110,11 +117,40 @@ func (cheBroker *ChePluginBroker) processPlugin(meta model.PluginMeta) error {
 	cheBroker.PrintDebug("Stared processing plugin '%s:%s'", meta.ID, meta.Version)
 	url := meta.URL
 
+	switch getTypeOfURL(url) {
+	case Archive:
+		return cheBroker.processArchive(&meta, url)
+	case Yaml:
+		return cheBroker.processYAML(&meta, url)
+	}
+	return nil
```
garagatyi: This line should not be reachable. Please, return error here.

brokers/che-plugin-broker/broker/broker.go
```diff
@@ -133,7 +169,7 @@ func (cheBroker *ChePluginBroker) processPlugin(meta model.PluginMeta) error {
 	}
 
 	cheBroker.PrintDebug("Resolving '%s:%s'", meta.ID, meta.Version)
-	err = cheBroker.resolveToolingConfig(&meta, pluginPath)
+	err = cheBroker.resolveToolingConfig(meta, pluginPath)
```
garagatyi: Can you elaborate on why this change is needed?

skabashnyuk: In the current branch, ```meta``` parameter declared in function as  ```processArchive(meta *model.PluginMeta, url string)``` in master it's declared as ```processPlugin(meta model.PluginMeta)```.

brokers/che-plugin-broker/broker/broker.go
garagatyi: abobe

brokers/che-plugin-broker/broker/broker.go
garagatyi: OK, thanks for the clarification

benoitf:
it looks strange that it has been merged based on the comments of the issue

slemeur:
Catching that after being on PTO.

Yes the use case looks odd to me. 
https://github.com/eclipse/che/issues/12246#issuecomment-449203200 is the way to go.

PR #26, opened 2019-01-25
Allow running brokers locally, add debug info

garagatyi:
### What does this PR do?
Now figuring out what went wrong with a plugin broker is quite painful.
This PR introduces changes that allow trying broker locally, debug it or so.
Changes:
- Add `-disable-push` option to brokers to run brokers without connecting to Workspace Master.
- Add printing of evaluated sidecar config to brokers right before exiting 
- Add examples of the configuration of brokers, so they can be easily tried locally.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,3 +4,5 @@
 */init/cmd/init-plugin-broker
 */vscode/cmd/vscode-extension-broker
 */cmd/cmd
+main
+*/cmd/config.json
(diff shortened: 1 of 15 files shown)


garagatyi:
@sleshchenko interesting idea. But this would involve cleaning up the code to remove not needed to debug logs. I didn't want to invest in that at this stage. Will consider as an improvement for future refactorings.

PR #28, opened 2019-01-30
Make Che 7 all fields camel-case

garagatyi:
### What does this PR do?
Make Che 7 all fields camel-case


metlos:
Answering my own question :) https://github.com/eclipse/che/pull/12560

PR #31, opened 2019-02-11
Add provisioning VS Code extension as Theia local plugin

garagatyi:
### What does this PR do?
Add provisioning VS Code extension as Theia local plugin.
When meta.yaml doesn't contain attribute `containerImage` VS Code extension is treated as local Theia plugin.

### DIFF
--- a/brokers/vscode/broker.go
+++ b/brokers/vscode/broker.go
@@ -21,7 +21,7 @@ import (
 	"path/filepath"
 	"regexp"
 
-	jsonrpc "github.com/eclipse/che-go-jsonrpc"
+	"github.com/eclipse/che-go-jsonrpc"
 	"github.com/eclipse/che-plugin-broker/brokers/theia"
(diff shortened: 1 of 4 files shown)


brokers/vscode/broker.go
```diff
@@ -135,9 +129,31 @@ func (b *Broker) processPlugin(meta model.PluginMeta) error {
 		}
 	}
 
+	image := meta.Attributes["containerImage"]
```
sleshchenko: Should not you check whether attributes are not nil as you did it before extension evaluation https://github.com/eclipse/che-plugin-broker/pull/31/files#diff-f9c03892980ba78514d5ef65ae08159eR100 ?

garagatyi: It is not needed anymore. If attributes are not present but URL is present we consider plugin local and hosted not in VS Code marketplace

PR #37, opened 2019-03-20
Create Unified plugin broker that merges functionality of all the plugin type specific brokers 

garagatyi:
### What does this PR do?
**Note:** This PR is designed to be reviewed in a commit-by-commit manner. Commit message of each commit explains what was changed in the corresponding commit.

Add new unified plugin broker that should co-exist with other plugin brokers for the time being.
It accepts plugin meta.yaml entries of all the types instead of type specific as other brokers do.
It sorts meta.yaml entries depending on the type and pass them to the appropriate broker implementations.
Tested locally with config.json containing all the meta.yaml entries from the repo for manual testing of other brokers.

Closes https://github.com/redhat-developer/rh-che/issues/1304


brokers/unified/broker.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/eclipse/che-plugin-broker/brokers/vscode"
+	"github.com/eclipse/che-plugin-broker/common"
+	"github.com/eclipse/che-plugin-broker/model"
+	"github.com/eclipse/che-plugin-broker/storage"
+	"github.com/eclipse/che-plugin-broker/utils"
+	"net/http"
+)
+
+const ChePluginType = "Che plugin"
+const EditorPluginType = "Che Editor"
```
sleshchenko: `Che Editor` but `Che plugin`, is it done by purpose?

garagatyi: It's a mistake. Thanks for catching

garagatyi:
@davidfestal not sure whether it is good time to do that. We also plan to move meta download from plugin registry in unified broker. With this improvement you would be able to remove some code from CRD POC codebase

garagatyi:
@davidfestal @sleshchenko @amisevsk I made plugin type matching case insensitive, please take a look and approve again

davidfestal:
> @davidfestal not sure whether it is good time to do that. We also plan to move meta download from plugin registry in unified broker. With this improvement you would be able to remove some code from CRD POC codebase

In fact I was already planning to open a very lightweight PR on plugin brokers with my local changes that allowed me to reuse them as is in the CRD Controller. I'll do it as soon as your current PR is merged: It included the promotion of the `processPlugin()` methods to `public` visibility, which is now provided by your PR.

And then I plan to integrate in the CRD controller new features / refactorings of the `che-plugin-broker` as soon as they are merged. The idea is to keep up-to-date and use standard components as much as possible.

garagatyi:
> If we're matching case-insensitive, does it make sense to include that in some tests? E.g. the type on the added test config could be lower case.

@amisevsk you are right, I'll need to add tests for that

garagatyi:
@amisevsk I added unit tests and couple of fixes that I was needed for testing or that I discovered along the way

brokers/che-plugin-broker/broker.go
```diff
@@ -36,29 +36,45 @@ const depFileBothLocationAndURLError = "Plugin dependency '%s:%s' contains both
...
 
 // ChePluginBroker is used to process Che plugins
-type ChePluginBroker struct {
+type ChePluginBroker interface {
+	Start([]model.PluginMeta)
+	PushEvents(tun *jsonrpc.Tunnel)
+	ProcessPlugin(meta model.PluginMeta) error
+}
+
+type chePluginBrokerImpl struct {
```
davidfestal: So it seems I'll have to rework my PR a bit more than expected, since I have made the `storage` field public in order to be able to retrieve the result from the broker after using `ProcessPlugin`.

davidfestal: Oh, or I would would now be able create a broker with a common external storage in which the results of all the brokers would be accumulated. Let me wait for you to merge, and I'll accomodate with your last changes.

garagatyi: It is kinda violation of encapsulation. If one uses field in tests it is bad but acceptable I suppose, but using it from other packages makes code too coupled. 
After this PR you can create brokers with provided storage as an argument to a method.
And frankly speaking you can use unified broker to process all the plugins and only then create kubernetes objects. It can improve quality of the code by decoupling different areas of responsibility.

garagatyi:
Guys I did some changes. I checked them on Che - everything works fine. I started workspace with plugins of each kind.
Please, add your reviews again

brokers/che-plugin-broker/broker.go
davidfestal: OK, I'll see that after you have merged. Thanks for the details.

PR #38, opened 2019-03-21
Allow using the `che-plugin-broker` as a library in the Workspace CRD controller

davidfestal:
### What does this PR do?

This PR introduces minimal changes to allow using the `che-plugin-broker` as a library in the Workspace CRD Controller Proof of concept.

Added 2 global configuration options to :
- allow skipping the copy of files in the `/plugins` directory (when we are only interested in generating the `ToolingConf` metadata)
- allow using `localhost` instead of the service name in the endpoint URL of the Theia or VSCode plugins


brokers/che-plugin-broker/broker.go
```diff
@@ -107,13 +107,13 @@ func (cheBroker *ChePluginBroker) PushEvents(tun *jsonrpc.Tunnel) {
 	cheBroker.Broker.PushEvents(tun, model.BrokerStatusEventType, model.BrokerResultEventType, model.BrokerLogEventType)
 }
 
-func (cheBroker *ChePluginBroker) processPlugin(meta model.PluginMeta) error {
+func (cheBroker *ChePluginBroker) ProcessPlugin(meta model.PluginMeta, onlyMetadata bool) error {
```
amisevsk: Semantically, I would prefer the "default" option to be `true` instead of false (i.e. rename property to something like `copyFiles`). This is more of a personal preference though -- up to you.

amisevsk: A further change here, to avoid requiring additional parameters to method calls, would be to break down the functions more directly, and have `ProcessPlugin` and `ProcessMetas` be separate entirely. The benefits there would be 
1. It seems like `onlyMetadata` applies only when the url is for an archive, which is maybe unclear
2. `onlyMetadata` causes early returns and has to be tracked across a few function calls

I also only mean this as a suggestion, though.

brokers/theia/broker.go
```diff
@@ -130,10 +130,10 @@ func (b *Broker) processPlugin(meta model.PluginMeta) error {
 	}
 	if pluginImage == "" {
 		// regular plugin
-		return b.injectTheiaFile(meta, archivePath)
+		return b.injectTheiaFile(meta, archivePath, onlyMetadata)
```
garagatyi: The purpose of this method is to inject file. It is better to refactor code instead of just adding boolean parameter to it. And with the remote plugin the problem is the same.

brokers/theia/sidecar.go
```diff
@@ -53,7 +54,11 @@ func AddExtension(toolingConf *model.ToolingConf, pj model.PackageJSON) {
 	sidecarEndpoint := toolingConf.Endpoints[0]
 	prettyID := re.ReplaceAllString(pj.Publisher+"_"+pj.Name, `_`)
 	sidecarTheiaEnvVarName := "THEIA_PLUGIN_REMOTE_ENDPOINT_" + prettyID
-	sidecarTheiaEnvVarValue := "ws://" + sidecarEndpoint.Name + ":" + strconv.Itoa(sidecarEndpoint.TargetPort)
+	sidecarHostname := sidecarEndpoint.Name
+	if cfg.UseLocalhostInPluginUrls {
```
garagatyi: When I was designing this utility it was supposed to not interact with config package. In fact, interacting with the config package is like using global variables.

brokers/vscode/broker.go
```diff
@@ -97,7 +97,7 @@ func (b *Broker) PushEvents(tun *jsonrpc.Tunnel) {
 	b.Broker.PushEvents(tun, model.BrokerStatusEventType, model.BrokerResultEventType, model.BrokerLogEventType)
 }
 
-func (b *Broker) processPlugin(meta model.PluginMeta) error {
+func (b *Broker) ProcessPlugin(meta model.PluginMeta, onlyMetadata bool) error {
```
garagatyi: It would interfere a lot with #37. I would suggest waiting for its merge.

brokers/che-plugin-broker/broker.go
davidfestal: I understand your remarks. Just 2 points:
- As for the `copyFiles` naming, my idea was more to point out that we might only be interested in the **metadata** part of things, just as mentioned in point 3 of this comment from @l0rd  https://github.com/eclipse/che/issues/12398#issue-398338347. But I can change it if necessary.
- As for splitting functions, I agree that some refactoring might be useful, but my idea was mainly to change the structure of the code as less as possible, knowing that a number of changes are ongoing or planned. I would have let the refactoring phase for the upcoming changes planned by @garagatyi.

brokers/theia/broker.go
davidfestal: I agree that some refactoring might be useful, but my idea was mainly to change the structure of the code as less as possible, knowing that a number of changes are ongoing or planned. I would have let the refactoring phase for the upcoming changes planned in the context of meta.yaml and plugin yaml merge.

brokers/theia/sidecar.go
davidfestal: That's what it is: a global variable that is taken into account every time we build an endpoint URL. BTW this global setting is only changed when the brokers are used as a library, so no change for your use-case.

brokers/vscode/broker.go
davidfestal: Do you mean because you did the exact same change in PR #37 (`processPlugin` to `ProcessPlugin`) ? 

I'm OK to wait for the merge of your PR, but do you have any idea if it will be merged soon ?

brokers/theia/sidecar.go
garagatyi: Here I'm more concerned about maintainability of the codebase rather than any particular case

brokers/vscode/broker.go
garagatyi: Yes, want to merge it today. Seems I finished everything and addressed all the reviews so far

davidfestal:
@amisevsk @ibuziuk @garagatyi @l0rd Could you review the PR again please ?

brokers/theia/sidecar.go
davidfestal: Maybe this concern could be tackled in the context of one of the future PRs that are still expected in the upcoming refactoring work on the `che-plugin-broker` ?

I tried to reduce the changes in this PR to the minimum, and in a way that introduce only local non-invasive changes.

PR #40, opened 2019-03-25
Make unified broker accept list of plugins ID:version s and download meta.yaml from plugin registry

amisevsk:
### What does this PR do?
Adds capability for the unified broker to process a list of plugin id, version, and registry fields and download the `meta.yaml`s directly from registry instead of expecting the metas to be present in config.

This PR is broken up into three commits:
1. Add functionality for processing plugin FQNs (fully qualified names) and downloading meta.yamls
2. Modify `ioutil` to enable testing
3. Add tests for ioutil.

### Design decisions
- Support both plugin metas and plugin FQNs (i.e. it is backward compatible)
- Switch between processing metas and FQNs is with argument `--download-metas`
- Default registry can be specified via argument `--registry-address`
  - Default registry is optional but only allowed when all plugin FQNs specify a registry
- Expected config file in FQN case is a json list of objects with fields `id`, `version` and optionally `registry`. This was chosen to mirror `PluginFQN` as used in Che, as the list of `PluginFQN` can be easily serialized into the configmap for the broker.

### Testing directly
This PR also adds a file `brokers/unified/cmd/config-plugin-ids.json`. To test directly, the command
```
cd ./brokers/unified/cmd
go build main.go
./main \
  --disable-push \
  --runtime-id wsId:env:ownerId \
  --registry-address https://che-plugin-registry.openshift.io/plugins \
  --metas ./config-plugin_ids.json \
  --download-metas
```

https://github.com/redhat-developer/rh-che/issues/1305

### DIFF
--- a/brokers/che-plugin-broker/broker.go
+++ b/brokers/che-plugin-broker/broker.go
@@ -23,12 +23,12 @@ import (
 
 	"gopkg.in/yaml.v2"
 
-	"github.com/eclipse/che-go-jsonrpc"
+	jsonrpc "github.com/eclipse/che-go-jsonrpc"
+	"github.com/eclipse/che-plugin-broker/cfg"
 	"github.com/eclipse/che-plugin-broker/common"
 	"github.com/eclipse/che-plugin-broker/model"
 	"github.com/eclipse/che-plugin-broker/storage"
 	"github.com/eclipse/che-plugin-broker/utils"
-	"github.com/eclipse/che-plugin-broker/cfg"
 )
 
 const pluginFileName = "che-plugin.yaml"
@@ -43,13 +43,6 @@ const (
 	Yaml
 )
 
-// ChePluginBroker is used to process Che plugins
-type ChePluginBroker interface {
-	Start([]model.PluginMeta)
-	PushEvents(tun *jsonrpc.Tunnel)
-	ProcessPlugin(meta model.PluginMeta) error
-}
-
 type chePluginBrokerImpl struct {
 	common.Broker
 	ioUtil  utils.IoUtil
@@ -57,7 +50,7 @@ type chePluginBrokerImpl struct {
 }
 
 // NewBroker creates Che plugin broker instance
-func NewBroker() ChePluginBroker {
+func NewBroker() common.BrokerImpl {
 	return &chePluginBrokerImpl{
 		common.NewBroker(),
 		utils.New(),
@@ -65,8 +58,8 @@ func NewBroker() ChePluginBroker {
 	}
 }
 
-// NewBroker creates Che plugin broker instance
-func NewBrokerWithParams(broker common.Broker, ioUtil utils.IoUtil, storage *storage.Storage) ChePluginBroker {
+// NewBrokerWithParams creates Che plugin broker instance
+func NewBrokerWithParams(broker common.Broker, ioUtil utils.IoUtil, storage *storage.Storage) common.BrokerImpl {
 	return &chePluginBrokerImpl{
 		Broker:  broker,
 		ioUtil:  ioUtil,
@@ -191,7 +184,7 @@ func (cheBroker *chePluginBrokerImpl) processArchive(meta *model.PluginMeta, url
 	if cfg.OnlyApplyMetadataActions {
 		return nil
 	}
-	
+
 	cheBroker.PrintDebug("Copying dependencies for '%s:%s'", meta.ID, meta.Version)
 	return cheBroker.copyDependencies(pluginPath)
 }
--- a/brokers/unified/broker.go
+++ b/brokers/unified/broker.go
@@ -37,9 +37,9 @@ type Broker struct {
 	common.Broker
 	Storage *storage.Storage
 
-	theiaBroker  theia.Broker
-	vscodeBroker vscode.Broker
-	cheBroker    broker.ChePluginBroker
+	theiaBroker  common.BrokerImpl
+	vscodeBroker common.BrokerImpl
+	cheBroker    common.BrokerImpl
 }
 
 // NewBroker creates Che broker instance
--- a/brokers/unified/broker.go
+++ b/brokers/unified/broker.go
@@ -15,16 +15,19 @@ package unified
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/eclipse/che-go-jsonrpc"
-	"github.com/eclipse/che-plugin-broker/brokers/che-plugin-broker"
+	"log"
+	"net/http"
+	"strings"
+
+	jsonrpc "github.com/eclipse/che-go-jsonrpc"
+	broker "github.com/eclipse/che-plugin-broker/brokers/che-plugin-broker"
 	"github.com/eclipse/che-plugin-broker/brokers/theia"
 	"github.com/eclipse/che-plugin-broker/brokers/vscode"
 	"github.com/eclipse/che-plugin-broker/common"
 	"github.com/eclipse/che-plugin-broker/model"
 	"github.com/eclipse/che-plugin-broker/storage"
 	"github.com/eclipse/che-plugin-broker/utils"
-	"net/http"
-	"strings"
+	"gopkg.in/yaml.v2"
 )
 
 const ChePluginType = "che plugin"
@@ -32,10 +35,15 @@ const EditorPluginType = "che editor"
 const TheiaPluginType = "theia plugin"
 const VscodePluginType = "vs code extension"
 
+// RegistryURLFormat specifies the format string for registry urls
+// when downloading metas
+const RegistryURLFormat = "%s/plugins/%s/%s/meta.yaml"
+
 // Broker is used to process Che plugins
 type Broker struct {
 	common.Broker
 	Storage *storage.Storage
+	utils   utils.IoUtil
 
 	theiaBroker  common.BrokerImpl
 	vscodeBroker common.BrokerImpl
@@ -54,13 +62,24 @@ func NewBroker() *Broker {
 	theiaBroker := theia.NewBrokerWithParams(commonBroker, ioUtils, storageObj, rand)
 	vscodeBroker := vscode.NewBrokerWithParams(commonBroker, ioUtils, storageObj, rand, httpClient)
 	return &Broker{
-		Storage: storageObj,
 		Broker:  commonBroker,
+		Storage: storageObj,
+		utils:   ioUtils,
 
 		theiaBroker:  theiaBroker,
 		vscodeBroker: vscodeBroker,
 		cheBroker:    cheBroker,
 	}
+}
+
+// DownloadMetasAndStart downloads metas from plugin registry for specified
+// pluginFQNs and then calls Start for those metas
+func (b *Broker) DownloadMetasAndStart(pluginFQNs []model.PluginFQN, defaultRegistry string) {
+	pluginMetas, err := b.getPluginMetas(pluginFQNs, defaultRegistry)
+	if err != nil {
+		b.PrintFatal("Failed to download plugin metas: %s", err)
+	}
+	b.Start(pluginMetas)
 }
 
 // Start executes plugins metas processing and sending data to Che master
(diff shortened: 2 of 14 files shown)


utils/ioutil.go
```diff
@@ -74,6 +96,55 @@ func (util *impl) Download(URL string, destPath string) error {
...
+			pluginURL = fmt.Sprintf("%s/%s/%s/meta.yaml", defaultRegistry, plugin.ID, plugin.Version)
+		}
+		pluginRaw, err := util.FetchJSON(pluginURL)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to fetch plugin meta.yaml from registry: %s", err)
+		}
+
+		var pluginMeta model.PluginMeta
+		if err := yaml.Unmarshal(pluginRaw, &pluginMeta); err != nil {
+			return nil, fmt.Errorf("Failed to unmarshal downloaded meta.yaml: %s", err)
```
sleshchenko: Please consider adding plugin id or even plugin URL to find broken meta.yaml easier.

brokers/unified/cmd/config-plugin-ids.json
```diff
@@ -0,0 +1,15 @@
...
+    "version": "0.1.17"
+  },
+  {
+    "id": "che-machine-exec-plugin",
+    "version": "0.0.1"
+  },
+  {
+    "id": "org.eclipse.che.vscode-redhat.java-and-xml",
+    "version": "0.38.0",
+    "registry": "https://raw.githubusercontent.com/eclipse/che-plugin-registry/shareSidecarWithDebug/plugins/"
```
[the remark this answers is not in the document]
garagatyi: :+1:

brokers/unified/cmd/main.go
```diff
@@ -34,6 +36,16 @@ func main() {
 		broker.PushEvents(statusTun)
 	}
 
-	metas := cfg.ReadConfig()
+	var metas []model.PluginMeta
+	if cfg.DownloadMetas {
+		pluginFQNs := cfg.ParsePluginFQNs()
+		pluginMetas, err := utils.New().GetPluginMetas(pluginFQNs, cfg.RegistryAddress)
+		if err != nil {
+			log.Fatalf("Failed to download plugin metas: %s", err)
```
garagatyi: Broker has special method for this purpose that insures that error message is sent to WS master.

cfg/cfg.go
```diff
@@ -101,6 +110,18 @@ func init() {
...
+		&RegistryAddress,
+		"registry-address",
+		"",
+		"Address of registry from which to retrieve meta.yamls when only plugin IDs are specified.",
+	)
+	flag.BoolVar(
+		&DownloadMetas,
+		"download-metas",
+		false,
+		"Download plugin metadata from registry instead of process already-downloaded metas.",
```
garagatyi: Other messages don't have trailing slashes. Please, align new ones

cfg/cfg.go
```diff
@@ -152,10 +173,35 @@ func Print() {
...
 // parses its content as array of Che plugin meta objects and returns it.
 // If any error occurs during read, log.Fatal is called.
+//
+// Deprecated
 func ReadConfig() []model.PluginMeta {
+	raw := readConfigFile()
+
+	metas := make([]model.PluginMeta, 0)
+	if err := json.Unmarshal(raw, &metas); err != nil {
+		log.Fatal(err)
```
garagatyi: Broker has special method for this purpose that insures that error message is sent to WS master. I see that we exit program here including in all codebase. 
Consider changing this file to return error when something happens and then main.go of brokers would be able to send error to WS master

model/model.go
```diff
@@ -68,6 +68,12 @@ type PluginMeta struct {
 	Extensions []string `json:"extensions" yaml:"extensions"`
 }
 
+type PluginFQN struct {
+	Registry string `json:"registry,omitempty" yaml:"registry,omitempty"`
```
garagatyi: Can you elaborate why we need omitempty here?

utils/ioutil.go
```diff
@@ -38,6 +42,8 @@ type IoUtil interface {
 	Unzip(arch string, dest string) error
 	Untar(tarPath string, dest string) error
 	CreateFile(file string, tr io.Reader) error
+	FetchJSON(url string) ([]byte, error)
+	GetPluginMetas(plugins []model.PluginFQN, defaultRegistry string) ([]model.PluginMeta, error)
```
garagatyi: Other methods in this file are business logic agnostic. Do you think we need to change the approach?

utils/ioutil.go
```diff
@@ -46,10 +46,26 @@ type IoUtil interface {
...
+// New creates an instance of IoUtil using the default http client.
 func New() IoUtil {
-	return &impl{}
+	return &impl{
+		httpClient: http.DefaultClient,
+	}
+}
+
+// NewWithClient creates an instance of IoUtil using the provided client for all requests.
+func NewWithClient(client HttpClient) IoUtil {
```
garagatyi: Can you elaborate why do we need this function?

brokers/unified/cmd/config-plugin-ids.json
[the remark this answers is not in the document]
amisevsk: Will do -- I copy-pasted from the issue :)

utils/ioutil.go
amisevsk: This method is used to mock HTTPClient in the IoUtil tests; it makes it easier to test request failure, etc and makes the tests more robust. See utils/mocks/MockHttpClient.go

model/model.go
amisevsk: "Need" is perhaps strong, but I included `omitempty` because I would expect `unmarshal -> marshal` to be idempotent, and we don't require the `registry` field in `PluginFQN` (i.e. from Che, it could be null and so omitted).

brokers/unified/cmd/config-plugin-ids.json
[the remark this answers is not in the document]
amisevsk: Changed to https://raw.githubusercontent.com/eclipse/che-plugin-registry/master/plugins/

cfg/cfg.go
```diff
@@ -101,6 +110,18 @@ func init() {
 		"Output events that are usually sent Che master instead of regular logs to imitate what a user can see."+
 			"`false` by default. Needed for testing and debugging purposes",
 	)
+	flag.StringVar(
+		&RegistryAddress,
+		"registry-address",
+		"",
+		"Address of registry from which to retrieve meta.yamls when only plugin IDs are specified.",
```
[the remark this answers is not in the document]
amisevsk: Updated message, good point :+1:

utils/ioutil.go
```diff
@@ -74,6 +96,55 @@ func (util *impl) Download(URL string, destPath string) error {
...
+func (util *impl) GetPluginMetas(plugins []model.PluginFQN, defaultRegistry string) ([]model.PluginMeta, error) {
+	metas := make([]model.PluginMeta, len(plugins))
+	for idx, plugin := range plugins {
+		log.Printf("Fetching plugin meta.yaml for %s:%s", plugin.ID, plugin.Version)
+		var pluginURL string
+		if plugin.Registry != "" {
+			pluginURL = fmt.Sprintf("%s/%s/%s/meta.yaml", plugin.Registry, plugin.ID, plugin.Version)
+		} else {
+			if defaultRegistry == "" {
+				return nil, errors.New("plugin does not specify registry and no default is provided")
```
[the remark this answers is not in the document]
amisevsk: go-lint suggests: `error strings should not be capitalized or end with punctuation or a newline`

Updated message to include plugin ID.

utils/ioutil.go
```diff
@@ -74,6 +96,55 @@ func (util *impl) Download(URL string, destPath string) error {
...
+			}
+			pluginURL = fmt.Sprintf("%s/%s/%s/meta.yaml", defaultRegistry, plugin.ID, plugin.Version)
+		}
+		pluginRaw, err := util.FetchJSON(pluginURL)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to fetch plugin meta.yaml from registry: %s", err)
```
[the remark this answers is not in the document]
amisevsk: :+1: Done

utils/ioutil.go
amisevsk: Added.

brokers/unified/cmd/main.go
amisevsk: I didn't realize -- done.

cfg/cfg.go
amisevsk: Not entirely sure what you mean, but I've removed `.` from the end of these messages.

cfg/cfg.go
amisevsk: This is a bit more involved of a change so it's broken out into a separate commit: https://github.com/eclipse/che-plugin-broker/pull/40/commits/b76456c6409d05e29010e1d0489228d717213949

In short: some of the broker implementations didn't include `common.Broker`, so to allow `broker.PrintFatal` to be called, I had to change interfaces and structs. Let me know what you think.

utils/ioutil.go
amisevsk: I agree about business logic in `ioutil` to an extent, but the method is fairly coupled with `fetchJSON`, which is definitely a `ioutil` method. Moving it elsewhere would require creating a `ioutil` instance, which I think is somewhat clunky. 

I'm open to suggestions though.

amisevsk: Also, at first glance, moving it out would make testing trickier, since mocking http.Client is necessary.

model/model.go
garagatyi: I see. Thanks for the explanation

garagatyi:
If you don't want to delete them, you need to avoid changes in them.

cfg/cfg.go
garagatyi: It is nice that we now send error to WS master properly. Though I didn't get why we need to rename  `PushEvents` function. And it is better to remove main.go in other brokers instead of extending their codebases.

utils/ioutil.go
garagatyi: We need to apply separation of concerns here. Util here is supposed to sugar cover boilerplate IO operations. Dealing with meta specific stuff is not supposed to be IO boilerplate code.
Thus I suggest moving code of downloading metas (or at least parsing them since we have download IO util function) to the broker. 
main.go of unified broker now also contains business logic and it wasn't supposed to.
Instead of that I would create `StartWithFQNs` function in addition to `Start` one and move business logic there. This should also help in avoiding changing other brokers (and you can avoid removing them as well).
Taking this into account main.go would do:
```go
if cfg.DownloadMetas {
    pluginFQNs, err := cfg.ParsePluginFQNs()
    if err ..... {}
    StartWithFQNs(pluginFQNs)
} else {
    pluginMetas, err := cfg.ReadConfig()
    if err ..... {}
    Start(pluginMetas)
}
```

Since we are going to remove `Start` function or make it do what `StartWithFQNs` does we should also deprecate it.

As for the unit testing VS Code broker tests contains cases with fetching something from remote server when remote server response mocking is needed. 
Here is function for that https://github.com/eclipse/che-plugin-broker/blob/b31e49ca87c87de4c251c7b0ffa6822682213612/brokers/test/utils.go#L139 that is used here https://github.com/eclipse/che-plugin-broker/blob/1451e5d1156bff4addb639050e848bd85969baca/brokers/vscode/broker_test.go#L65 in conjunction with https://github.com/eclipse/che-plugin-broker/blob/1451e5d1156bff4addb639050e848bd85969baca/brokers/vscode/broker_test.go#L877

utils/ioutil.go
garagatyi: There is a simpler mocking technique. See explanation how it is used in VS Code broker tests in one of my other comments

garagatyi:
@amisevsk I'm working on another issue that is related to this PR. Do you have draft changes on WS master side to use broker with these changes? I am asking because you could have them to test your changes integrated in Che and it would be better if I do not overlap with your changes or would be able to identify if those changes I wanted to do are compatible with yours

utils/ioutil.go
amisevsk: :+1: Makes sense, will do.

amisevsk:
> Do you have draft changes on WS master side to use broker with these changes?

I don't yet, but should have something today.

utils/ioutil.go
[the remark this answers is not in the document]
amisevsk: (The reasoning behind this is that often error messages are combined: `fmt.Sprintf("could not do x: %s", err)`, so adding capitalization introduces caps halfway through the message.)

cfg/cfg.go
garagatyi: I decided to remove them separately as this is left over from my last PR rather then something your PR specifically needs

amisevsk: Thanks for that -- I'll rebase onto current master.

amisevsk: Regarding the `PushEvents` function, that is something I'm currently working on a decent solution to; in e.g. the Theia broker, we're defining a `Broker` interface that includes a `PushEvents` function. To be able to use a Theia broker's `PrintFatal` we need to include `common.Broker` in this interface, which causes a name collision on the `PushEvents` function. However, since we're no longer using the non-unified brokers standalone, this can likely be removed.

amisevsk:
@garagatyi I've rebased onto master and added two commits:

- https://github.com/eclipse/che-plugin-broker/pull/40/commits/7ebbb2874fecfd6fdadcb2282b00765951e17609 is a minor refactor to move the broker interface to the common package (since there are three identical interfaces in each of che plugin, theia, and vscode brokers)

- https://github.com/eclipse/che-plugin-broker/pull/40/commits/f9f828a72f09af51dd19180a5d41fc801f4871a5 does the requested changes: it
  - Moves meta downloading to unified broker.
  - Reworks tests as requested.

PTAL

brokers/unified/broker.go
```diff
@@ -125,10 +141,41 @@ func (b *Broker) ProcessPlugins(metas []model.PluginMeta) error {
...
+		var registry string
+		if plugin.Registry != "" {
+			registry = strings.TrimSuffix(plugin.Registry, "/")
+		} else {
+			if defaultRegistry == "" {
+				return nil, fmt.Errorf("plugin '%s' does not specify registry and no default is provided", plugin.ID)
+			}
+			registry = strings.TrimSuffix(defaultRegistry, "/")
+		}
+		pluginURL := fmt.Sprintf("%s/%s/%s/meta.yaml", registry, plugin.ID, plugin.Version)
```
garagatyi: Taking into account that Che properties specify plugin registry without `/plugins` suffix (`che.workspace.plugin_registry_url=https://che-plugin-registry.openshift.io`) it would be better to align them and add `/plugins` here too. Treat it as API, not just file hosting.

brokers/che-plugin-broker/broker.go
```diff
@@ -43,30 +43,23 @@ const (
 	Yaml
 )
 
-// ChePluginBroker is used to process Che plugins
-type ChePluginBroker interface {
```
garagatyi: @davidfestal this is something that can affect your work

brokers/unified/broker.go
garagatyi: Since in a case of old plugin `/plugins` can be treated as `publisher` we probbaly can not avoid setting plugin registry without `/plugins`

brokers/che-plugin-broker/broker.go
amisevsk: Since they're all the same, you should be able to use `common.BrokerImpl` instead.

brokers/unified/broker.go
amisevsk: I've pulled the URL format into a const at the top, but now I'm thinking whether it makes sense to include it in the config at some point. There would also be the problem of using multiple registries for different plugins; they would all have to conform to our current spec which could be tricky.

PR #42, opened 2019-03-27
Cleanup brokers

garagatyi:
### What does this PR do?
Remove main.go of brokers that are no longer in use.
Remove not needed anymore dockerfiles and other related stuff.
Adapt docs after merging brokers functionality.

https://github.com/eclipse/che-plugin-broker/pull/37


brokers/che-plugin-broker/cmd/config.json
```diff
@@ -1,32 +0,0 @@
...
-    "name": "Che Samples REST API Sidecar Plugin",
-    "title": "Che Samples REST API Sidecar Plugin",
-    "description": "Che Plug-in with Theia plug-in and container definition providing a service",
-    "icon": "https://www.eclipse.org/che/images/ico/16x16.png",
-    "url": "https://github.com/ws-skeleton/che-service-plugin/releases/download/untagged-5c4c888ff2de8ae7a5e2/che-service-plugin.tar.gz"
-},
-{
-    "id": "org.eclipse.che.editor.theia",
-    "version": "1.0.0",
-    "type": "Che Editor",
```
sleshchenko: Was it done by a purpose that this config contains two Che Theia editors?

brokers/vscode/cmd/manual-testing-several-extensions-in-one-plugin.json
```diff
@@ -1,41 +0,0 @@
-[
-  {
-    "id": "org.eclipse.che.vscode-redhat.java-and-yaml-in-sidecar",
```
sleshchenko: Name of this file gives pretty clear what is expected during manual testing. I see that now `config.json` for unified broker contains all plugins with all plugins types and it is not clear what to expect as a result of work. It may be an issue to check result of unified broker work if there are so much to do, like check that sidecar is generated for Theia Remote Plugins, multiple vscode extensions are run in the same sidecar. Maybe it would be better to split `config.json` into separated files?
BTW It is not really related to this PR but just proposal for further improvement of unified Broker.

brokers/che-plugin-broker/cmd/config.json
garagatyi: Not really. It checked that it is possible to use either tar.gz or raw link to YAML

brokers/vscode/cmd/manual-testing-several-extensions-in-one-plugin.json
garagatyi: Yeah, I was thinking about writing some integration tests. Some case are trickier to test though.

PR #48, opened 2019-04-15
Add release script

garagatyi:
### What does this PR do?
Release script adds git tag and pushes containers images to Dockerhub.
Note that this script creates ugly entry like this https://github.com/eclipse/che-plugin-broker/releases/tag/v0.15.1
But we can create nice looking release like this https://github.com/eclipse/che-plugin-broker/releases/tag/v0.15.2 manually on Github repo release view https://github.com/eclipse/che-plugin-broker/releases

### DIFF
--- a/.ci/release.sh
+++ b/.ci/release.sh
@@ -2,6 +2,15 @@
 
 set -e
 
+if [ -z "${GITHUB_TAG}" ]; then
+  echo "Variable GITHUB_TAG is missing"
+  exit 1
+fi
+if [ -z "${IMAGE_TAG}" ]; then
+  echo "Variable IMAGE_TAG is missing"
+  exit 1
+fi
+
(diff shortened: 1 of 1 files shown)


.ci/release.sh
```diff
@@ -0,0 +1,25 @@
...
+
+# create and push new tag
+git tag $GITHUB_TAG
+git push origin $GITHUB_TAG
+
+# checkout to new tag
+git checkout $GITHUB_TAG
+
+# Unified Plugin Broker
+docker build -t eclipse/che-unified-plugin-broker:${IMAGE_TAG} -f ${ROOT_DIR}/build/unified/Dockerfile ${ROOT_DIR}
```
riuvshin: I would add checks for `IMAGE_TAG` and `GITHUB_TAG ` and fail if these are not set, this will protect this script from being executed locally with not ENVs set

garagatyi: Done

PR #49, opened 2019-04-19
Naming convention

garagatyi:
### What does this PR do?
Make broker download metas using new naming convention.
Details at eclipse/che-plugin-registry#55
Cleanup not needed code.

Related to eclipse/che-plugin-registry#55

### DIFF
--- a/cfg/cfg.go
+++ b/cfg/cfg.go
@@ -62,11 +62,6 @@ var (
 	// RegistryAddress address of the plugin registry, if plugin IDs are specified in config instead of metas.
 	// Used as a default registry if a plugin fully-qualified name does not specify a registry.
 	RegistryAddress string
-
-	// DownloadMetas specifies whether the broker should download plugin metas from the registry. If
-	// true, then config file should be a list of plugin fully-qualified names. Otherwise, the config
-	// file should contain the plugin metas to be processed.
-	DownloadMetas bool
 )
 
 func init() {
@@ -126,12 +121,6 @@ func init() {
 		"",
 		"Default address of registry from which to retrieve meta.yamls when plugin FQNs do not specify a registry. Ignored unless --download-metas is set",
 	)
-	flag.BoolVar(
-		&DownloadMetas,
-		"download-metas",
-		false,
-		"Download plugin metadata from registry instead of process already-downloaded metas",
-	)
 }
 
 // Parse parses configuration.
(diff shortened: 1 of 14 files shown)


cfg/cfg.go
```diff
@@ -62,11 +62,6 @@ var (
 	// RegistryAddress address of the plugin registry, if plugin IDs are specified in config instead of metas.
 	// Used as a default registry if a plugin fully-qualified name does not specify a registry.
 	RegistryAddress string
-
-	// DownloadMetas specifies whether the broker should download plugin metas from the registry. If
-	// true, then config file should be a list of plugin fully-qualified names. Otherwise, the config
-	// file should contain the plugin metas to be processed.
-	DownloadMetas bool
```
sleshchenko: It was needed for backward compatibility and not it is not needed anymore, right?

garagatyi: Yes

cfg/cfg.go
```diff
@@ -126,12 +121,6 @@ func init() {
 		"",
 		"Default address of registry from which to retrieve meta.yamls when plugin FQNs do not specify a registry. Ignored unless --download-metas is set",
 	)
-	flag.BoolVar(
-		&DownloadMetas,
-		"download-metas",
-		false,
-		"Download plugin metadata from registry instead of process already-downloaded metas",
-	)
```
amisevsk: Removing this now would result in a minor version bump (0.16.x) and require changes in upstream Che.

cfg/cfg.go
```diff
@@ -126,12 +121,6 @@ func init() {
 		"",
 		"Default address of registry from which to retrieve meta.yamls when plugin FQNs do not specify a registry. Ignored unless --download-metas is set",
```
amisevsk: Please update this message as well, since `--download-metas` is being removed.

amisevsk:
Could you also remove `--download-metas` from the [Makefile](https://github.com/eclipse/che-plugin-broker/blob/e93ad4de33e4e07d4e7130e66faf69e937b944c7/Makefile#L48)?

cfg/cfg.go
garagatyi: Yes, but I'm considering change version to `v1.0.0` since it is backward incompatible and first number is supposed be reflect such changes. WDYT?

cfg/cfg.go
garagatyi: nice catch

cfg/cfg.go
amisevsk: I'm open to either option but worry we're not to the point of calling it a `v1.0` release, especially with the various naming convention/registry changes still coming down the pipe. It's still okay to break backward compatibility in version zero so I'm leaning more in that direction.

PR #50, opened 2019-04-23
Fix usage of custom registry

garagatyi:
### What does this PR do?
/plugins needs to be added to the default plugin registry only.
This makes broker code compatible with meta downloading that was
previously residing on Che wsmaster side.

Related to https://github.com/eclipse/che/issues/13208


amisevsk:
Thanks @garagatyi, I should have caught this. We should look into more robustly specifying registry URL format down the line, since our current way might be a little fragile.

garagatyi:
Yeah, we should just be able to specify a plain URL to content

PR #56, opened 2019-05-03
[WIP] Use `localhost` in plugin endpoints by default

davidfestal:
### What does this PR do?

Since, in the current Che implementation, **all plugin containers are run on the same POD** as the Theia IDE container, let's change the default behavior of the plugin brokers to **use `localhost` as the IP of side-car plugin endpoints**, instead of using the short service name, which, in some cases, cannot be reached.

https://github.com/eclipse/che/issues/11018

### Has this PR been tested ?

Yes, on Minishift Che Addon with the v2 prod-preview plugin registry, the nightly che-server docker image, and the modified unified plugin broker.

### DIFF
--- a/brokers/theia/sidecar.go
+++ b/brokers/theia/sidecar.go
@@ -31,7 +31,9 @@ func GenerateSidecarTooling(image string, pj model.PackageJSON, rand common.Rand
 	endpoint := generateTheiaSidecarEndpoint(rand)
 	setEndpoint(tooling, endpoint)
 	AddExtension(tooling, pj)
-
+  if cfg.UseLocalhostInPluginUrls {
+		tooling.Endpoints = tooling.Endpoints[1:]
+	}
 	return tooling
 }
 
--- a/cfg/cfg.go
+++ b/cfg/cfg.go
@@ -53,7 +53,8 @@ var (
 	// UseLocalhostInPluginUrls configures the broker to use the `localhost` name
 	// instead of the Kubernetes service name to build Theia or VSCode plugin
 	// endpoint URL
-	UseLocalhostInPluginUrls bool
+	// True by default, at least as long as all the plugin containers are in the same POD
+	UseLocalhostInPluginUrls = true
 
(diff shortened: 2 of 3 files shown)


cfg/cfg.go
```diff
@@ -53,7 +53,8 @@ var (
 	// UseLocalhostInPluginUrls configures the broker to use the `localhost` name
 	// instead of the Kubernetes service name to build Theia or VSCode plugin
 	// endpoint URL
-	UseLocalhostInPluginUrls bool
+	// True by default, at least as long as all the plugin containers are in the same POD
+	UseLocalhostInPluginUrls = true
```
garagatyi: This means that we have no way to disable this mode, but still have a lot of code related to this. Why not add an option to disable/enable it (even though it can be enabled by default) OR remove not needed anymore code?

davidfestal: > This means that we have no way to disable this mode

We can disable it when using the broker as a library still

> Why not add an option to disable/enable it 

I could add an option yes, but that didn't seem that necessary and urgent (since there's no use-case for this in Che presently), compared to doing the simplest and smallest change possible.

> remove not needed anymore code

This code would still be useful if, in the future, we decide to add plugins in distinct pods.

davidfestal:
> Tests are not passing

@garagatyi I'll fix tests as soon as your last PRs are merged, since I expect a number of conflicts to occur.

brokers/theia/sidecar.go
```diff
@@ -31,7 +31,9 @@ func GenerateSidecarTooling(image string, pj model.PackageJSON, rand common.Rand
 	endpoint := generateTheiaSidecarEndpoint(rand)
 	setEndpoint(tooling, endpoint)
 	AddExtension(tooling, pj)
-
+  if cfg.UseLocalhostInPluginUrls {
```
amisevsk: Indentation issue here -- spaces instead of tabs

davidfestal:
A new PR will be opened from another branch after rebasing on the last master.

PR #58, opened 2019-05-10
Avoid the need to set userland-proxy

davidfestal:
### What does this PR do?

Since, in the current Che implementation, **all plugin containers are run on the same POD** as the Theia IDE container, let's change the default behavior of the plugin brokers to **use `localhost` as the IP of side-car plugin endpoints**, instead of using the short service name, which, in some cases, cannot be reached.

https://github.com/eclipse/che/issues/11018

This should also fix issue https://github.com/redhat-developer/rh-che/issues/1371 as mentionned in 
https://github.com/redhat-developer/rh-che/issues/1371#issuecomment-484504135

### Has this PR been tested ?

Tested on upstream Che on minishift, with the docker image corresponding to the (now merged) PR https://github.com/eclipse/che/pull/13311


.vscode/launch.json
```diff
@@ -0,0 +1,24 @@
+{
```
amisevsk: This file should be in `.gitignore`, could you remove it and add `.vscode` to the ignore list?

davidfestal: why ? isn't this file useful, to allow debugging the brokers ?

amisevsk: It's `.vscode` specific though. So far we've preferred editor-agnostic utils for testing, such as the [`Makefile`](https://github.com/eclipse/che-plugin-broker/blob/master/Makefile)

cfg/cfg.go
```diff
@@ -115,6 +117,13 @@ func init() {
 		"Output events that are usually sent Che master instead of regular logs to imitate what a user can see."+
 			"`false` by default. Needed for testing and debugging purposes",
 	)
+	flag.BoolVar(
+		&UseLocalhostInPluginUrls,
+		"use-localhost-in-plugin-urls",
+		true,
+		"This configures the broker to use the `localhost` name instead of the Kubernetes service name to build Theia or VSCode plugin endpoint URL."+
+			"`true` by default since until now all remote VSCode or Theia plugin containers are started on the same POD as the Theia IDE container",
```
benoitf:
```suggestion
			"`true` by default since until now all remote VS Code or Theia plugin containers are started on the same POD as the Theia IDE container",
```

cfg/cfg.go
```diff
@@ -53,7 +53,9 @@ var (
 	// UseLocalhostInPluginUrls configures the broker to use the `localhost` name
 	// instead of the Kubernetes service name to build Theia or VSCode plugin
 	// endpoint URL
-	UseLocalhostInPluginUrls bool
+	// True by default since until now all remote VSCode or Theia plugin containers
```
benoitf:
```suggestion
	// True by default since until now all remote VS Code or Theia plugin containers
```

.vscode/launch.json
davidfestal: I added `.vscode/` in `.gitignore` and remove the `launch.json` vscode-specific file.

davidfestal:
@amisevsk This might be interesting to create a minor version of the v17 release, no ? So that it would be used by upstream master ? at least to do it as soon as prod-preview will use the v3 plugin registry (if not already the case). wdyt ?

amisevsk:
@davidfestal `che-plugin-broker:v0.17.1` is released.

PR #59, opened 2019-05-11
Add ability to use plugin reference

mshaposhnik:
### What does this PR do?
This is a PR containing plugin broker part of adding possibility to define chePlugin/cheEditor type components using absolute reference to plugin meta or using custom registryUrl-s.

Fixes Introduce registryUrl field for cheEditor/chePlugin components https://github.com/eclipse/che/issues/13104

### DIFF
--- a/brokers/unified/broker.go
+++ b/brokers/unified/broker.go
@@ -260,7 +260,7 @@ func sortMetas(metas []model.PluginMeta) (che []model.PluginMeta, vscode []model
 func getRegistryURL(plugin model.PluginFQN, defaultRegistry string) (string, error) {
 	var registry string
 	if plugin.Registry != "" {
-		registry = strings.TrimSuffix(plugin.Registry, "/")
+		registry = strings.TrimSuffix(plugin.Registry, "/") + "/plugins"
 	} else {
 		if defaultRegistry == "" {
 			return "", fmt.Errorf("plugin '%s' does not specify registry and no default is provided", plugin.ID)
--- a/brokers/unified/broker.go
+++ b/brokers/unified/broker.go
@@ -181,22 +181,29 @@ func validateMetas(metas []model.PluginMeta) error {
 func (b *Broker) getPluginMetas(plugins []model.PluginFQN, defaultRegistry string) ([]model.PluginMeta, error) {
 	metas := make([]model.PluginMeta, 0, len(plugins))
 	for _, plugin := range plugins {
-		log.Printf("Fetching plugin meta.yaml for %s", plugin.ID)
-		registry, err := getRegistryURL(plugin, defaultRegistry)
-		if err != nil {
-			return nil, err
+		var pluginURL string
+		if plugin.Reference != "" {
+			pluginURL = plugin.Reference
+			log.Printf("Fetching plugin meta.yaml from reference %s", pluginURL)
+		} else {
+			log.Printf("Fetching plugin meta.yaml for %s", plugin.ID)
+			registry, err := getRegistryURL(plugin, defaultRegistry)
+			if err != nil {
+				return nil, err
+			}
+			pluginURL = fmt.Sprintf(RegistryURLFormat, registry, plugin.ID)
 		}
-		pluginURL := fmt.Sprintf(RegistryURLFormat, registry, plugin.ID)
+
 		pluginRaw, err := b.utils.Fetch(pluginURL)
 		if err != nil {
 			if httpErr, ok := err.(*utils.HTTPError); ok {
 				return nil, fmt.Errorf(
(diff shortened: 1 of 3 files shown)


brokers/unified/broker.go
```diff
@@ -253,7 +260,7 @@ func sortMetas(metas []model.PluginMeta) (che []model.PluginMeta, vscode []model
 func getRegistryURL(plugin model.PluginFQN, defaultRegistry string) (string, error) {
 	var registry string
 	if plugin.Registry != "" {
-		registry = strings.TrimSuffix(plugin.Registry, "/")
+		registry = strings.TrimSuffix(plugin.Registry, "/") + "/plugins"
```
amisevsk: Adding `/plugins` should be unnecessary, no? I'd vote against modifying user input.

mshaposhnik: Agree, that can be set explicitly if needed.

sleshchenko: I think it's more question to Che Plugin Registry, 
@benoitf suggested to make root return README.md, and `/plugins/`->  `index.json`.
Adding `plugins` is a kind of Plugin Registry API. If we stop using `plugins`, then I think it's needed to rework Plugin Registry and make root request return `index.json`. It's my personal feeling and it may be wrong.

About settings `plugins` explicitly by user:
it may be OK but ATM it's not clear enough like a user deploys his custom plugin registry based on our, retrieves Plugin Registry Route URL from OpenShift console and it's logical to use it as `registryUrl` without `/plugins` segment.

If we change this behavior, then as Alex commented https://github.com/eclipse/che/issues/13104#issuecomment-483245433, we should consider making the corresponding changes for `UD`, `WsMaster`  and `Theia`.

model/model.go
```diff
@@ -92,7 +93,7 @@ type Container struct {
 	Name         string        `json:"name,omitempty" yaml:"name,omitempty"`
 	Image        string        `json:"image,omitempty" yaml:"image,omitempty"`
 	Env          []EnvVar      `json:"env" yaml:"env"`
-	Commands     []Command     `json:"commands" yaml:"commands"`
+	Commands     []Command     `json:"editorCommands" yaml:"editorCommands"`
```
sleshchenko: I guess it should be reverted.

skabashnyuk:
LGTM

brokers/unified/broker_test.go
```diff
@@ -1039,7 +1039,7 @@ func TestBroker_getPluginMetas(t *testing.T) {
 				errRegexp: nil,
 				fetchURL: fmt.Sprintf(
 					RegistryURLFormat,
-					pluginFQNWithRegistry.Registry,
+					pluginFQNWithRegistry.Registry+"/plugins",
```
sleshchenko: It looks unformatted. Please fix it manually or execute `gofmt`

brokers/unified/broker.go
```diff
@@ -191,22 +191,28 @@ func validateMetas(metas []model.PluginMeta) error {
...
-	log.Printf("Fetching plugin meta.yaml for %s", plugin.ID)
-	registry, err := getRegistryURL(plugin, defaultRegistry)
-	if err != nil {
-		return nil, err
+	var pluginURL string
+	if plugin.Reference != "" {
+		pluginURL = plugin.Reference
+		log.Printf("Fetching plugin meta.yaml from reference %s", pluginURL)
+	} else {
+		log.Printf("Fetching plugin meta.yaml for %s", plugin.ID)
```
sleshchenko: I think it makes sense to log registry URL here as well.
Or have single log for both cases (reference, registry): `Fetching plugin meta.yaml from %s`, pluginURL

brokers/unified/broker_test.go
```diff
@@ -1066,7 +1066,7 @@ func TestBroker_getPluginMetas(t *testing.T) {
 				defaultRegistry: defaultRegistry,
 			},
 			want: want{
-				errRegexp: regexp.MustCompile("failed to fetch plugin meta.yaml for plugin .* from registry .*"),
+				errRegexp: regexp.MustCompile("failed to fetch plugin meta.yaml from URL .*"),
```
sleshchenko: Please consider adding test case for new use-case with `reference` specified

brokers/unified/broker_test.go
mshaposhnik: gofmt fixes nothing for me in this place

brokers/unified/broker.go
mshaposhnik: makes sense

brokers/unified/broker_test.go
mshaposhnik: added

PR #60, opened 2019-05-13
Compatibility the new automatic multicast discovery and plugin file isolation

davidfestal:
### What does this PR do?

This PR Implements the compatibility the new automatic multicast discovery and the isolation of plugin files (work done on `che-theia` by @benoitf in PR https://github.com/eclipse/che-theia/pull/91 

This fixes issue https://github.com/redhat-developer/rh-che/issues/1354 (also described in https://github.com/eclipse/che/issues/13272 and https://github.com/eclipse/che/issues/12395#issuecomment-470591054

This also includes a fix for issue https://github.com/eclipse/che/issues/13349

### Was this PR tested ?

Yes, with upstream Che `nightly` docker image running on Minishift.
Still want to test the che-in-che flow  (Development and test of remote Theia plugins under Che-Theia)


brokers/unified/vscode/broker.go
```diff
@@ -245,8 +245,9 @@ func extensionOrURL(extensionOrURL string) (extension string, URL string) {
 	}
 }
 
-func (b *brokerImpl) generatePluginArchiveName(plugin model.ChePlugin) string {
-	return fmt.Sprintf("%s.%s.%s.%s", plugin.Publisher, plugin.Name, plugin.Version, b.rand.String(10))
+func (b *brokerImpl) generatePluginArchiveName(plugin model.ChePlugin, archivePath string) string {
```
amisevsk: This likely won't work on current master; `archivePath` is generated and no longer resolves filename, and resolving filename from URL won't work for many URLs.

amisevsk: Ah nevermind I see this commit also modifies how `archivePath` is generated.

I worry that depending on URLs ending in the filename will cause problems -- e.g. https://marketplace.visualstudio.com/_apis/public/gallery/publishers/vscjava/vsextensions/vscode-java-debug/0.16.0/vspackage is a valid URL.

davidfestal: This is fixed with commit https://github.com/eclipse/che-plugin-broker/pull/60/commits/ce4c699d7b92f7fdbbac33bfaee10e59532bcc8b

utils/ioutil.go
```diff
@@ -55,29 +58,52 @@ func New() IoUtil {
...
+			if _, params, err := mime.ParseMediaType(dispo); err == nil {
+				contentDispoFilename = params["filename"]
+			}
+			if strings.HasSuffix(contentDispoFilename, "/") || strings.Contains(contentDispoFilename, "\x00") {
+				contentDispoFilename = ""
+			}
+			contentDispoFilename = path.Base(path.Clean("/" + contentDispoFilename))
+			if contentDispoFilename == "." || contentDispoFilename == "/" {
+				contentDispoFilename = ""
+			}
```
amisevsk: Could you add a few comments about why the steps above are necessary? Having multiple cases where `contentDispoFilename` is reset to the empty string is a little confusing. Some examples of cases being addressed would be useful.

davidfestal: I refactored a bit the code to make it clearer in commit  https://github.com/eclipse/che-plugin-broker/pull/60/commits/103113f16e51cf895a242e69ec1659ebc3b82f53

Can you tell if you think a comment is still necessary ?

davidfestal:
@ibuziuk @amisevsk @benoitf I updated the ReadMe accordingly. Do you want to have a look before we merge ?

PR #61, opened 2019-05-17
Make broker preserve extensions for downloaded plugins

amisevsk:
### What does this PR do?
- Add function `DownloadPreserveFilename` to `ioutil`
  - Attempts to use Content-Disposition header to determine filename
  - If unable, behaves as previous `Download()` function
  - Necessary for keeping file extensions (e.g. `.theia` for theia
    plugins)
- Make broker add file extension, if available, to copied local plugins
  - This is necessary as files without `.theia` extension are not picked
    up as plugins at workspace runtime
- Remove mentions of VS Code in logging when step can apply to Theia
  plugins as well
- `gofmt` ran automatically on edited files so there are some formatting
  changes.

https://github.com/eclipse/che/issues/13349

### DIFF
--- a/brokers/unified/vscode/broker.go
+++ b/brokers/unified/vscode/broker.go
@@ -49,7 +49,7 @@ type brokerImpl struct {
 	localhostSidecar bool
 }
 
-// NewBrokerWithParams creates Che VS Code extension broker instance
+// NewBrokerWithParams creates Che VS Code/Theia extension broker instance
 func NewBrokerWithParams(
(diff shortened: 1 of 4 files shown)


amisevsk:
cc: @vitaliy-guliy 

Built PR is available on dockerhub at `amisevsk/che-unified-plugin-broker:dev`

amisevsk:
Closing as it will be implemented as part of #60 (which makes other changes to VS Code/Theia plugin processing.

PR #62, opened 2019-05-21
Make the brokers usable as a library again

davidfestal:
### What does this PR do?

This PR slightly refactors the code in order to enable other components to use brokers as a GO library.
This is useful for the new Che Workspace Operator that embeds brokering, and was broken by the last `che-plugin-broker` changes.

No issue.

### DIFF
--- a/brokers/unified/broker.go
+++ b/brokers/unified/broker.go
@@ -48,20 +48,25 @@ type Broker struct {
 }
 
 // NewBroker creates Che broker instance
-func NewBroker(localhostSidecar bool) *Broker {
-	commonBroker := common.NewBroker()
-	ioUtils := utils.New()
-	storageObj := storage.New()
-	httpClient := &http.Client{}
-	rand := common.NewRand()
-
-	vscodeBroker := vscode.NewBrokerWithParams(commonBroker, ioUtils, storageObj, rand, httpClient, localhostSidecar)
-	return &Broker{
-		Broker:       commonBroker,
-		Storage:      storageObj,
-		utils:        ioUtils,
-		vscodeBroker: vscodeBroker,
+func NewBrokerWithParams(
+	commonBroker common.Broker,
+	ioUtil utils.IoUtil,
+	storage storage.Storage,
+	rand common.Random,
+	httpClient *http.Client,
+	localhostSidecar bool) *Broker {
+		vscodeBroker := vscode.NewBrokerWithParams(commonBroker, ioUtil, storage, rand, httpClient, localhostSidecar)
+		return &Broker{
+			Broker:       commonBroker,
+			Storage:      storage,
+			utils:        ioUtil,
+			vscodeBroker: vscodeBroker,
+		}
 	}
+
+// NewBroker creates Che broker instance
+func NewBroker(localhostSidecar bool) *Broker {
+	return NewBrokerWithParams(common.NewBroker(), utils.New(), storage.New(), common.NewRand(), &http.Client{}, localhostSidecar)
 }
 
 // DownloadMetasAndStart downloads metas from plugin registry for specified
@@ -119,7 +124,7 @@ func (b *Broker) ProcessPlugins(metas []model.PluginMeta) error {
 
 	b.PrintInfo("Starting Che plugins and editor processing")
 	for _, meta := range cheMetas {
-		plugin := convertMetaToPlugin(meta)
+		plugin := ConvertMetaToPlugin(meta)
 		err = b.Storage.AddPlugin(plugin)
 		if err != nil {
 			return err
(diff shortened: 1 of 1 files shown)


brokers/unified/broker.go
```diff
@@ -48,20 +48,25 @@ type Broker struct {
...
-	httpClient := &http.Client{}
-	rand := common.NewRand()
-
-	vscodeBroker := vscode.NewBrokerWithParams(commonBroker, ioUtils, storageObj, rand, httpClient, localhostSidecar)
-	return &Broker{
-		Broker:       commonBroker,
-		Storage:      storageObj,
-		utils:        ioUtils,
-		vscodeBroker: vscodeBroker,
+func NewBrokerWithParams(
```
amisevsk: Please update the godoc comment (first word should match function name)

brokers/unified/broker.go
```diff
@@ -137,42 +142,84 @@ func (b *Broker) ProcessPlugins(metas []model.PluginMeta) error {
 	return nil
 }
 
+func ValidateMeta(meta model.PluginMeta) error {
+	switch meta.APIVersion {
+	case "":
+		return fmt.Errorf("Plugin '%s' is invalid. Field 'apiVersion' must be present", meta.ID)
```
amisevsk: Minor style detail: error messages should start with a lowercase letter, per https://github.com/golang/go/wiki/CodeReviewComments#error-strings

brokers/unified/broker.go
```diff
@@ -137,42 +142,84 @@ func (b *Broker) ProcessPlugins(metas []model.PluginMeta) error {
 	return nil
 }
 
+func ValidateMeta(meta model.PluginMeta) error {
```
amisevsk: It's a little strange to me to make `ValidateMeta` an exported function but keep `validateMetas` internal. Is there a reason for this?

brokers/unified/broker.go
```diff
@@ -137,42 +142,84 @@ func (b *Broker) ProcessPlugins(metas []model.PluginMeta) error {
...
-			if len(meta.Spec.Containers) > 1 {
-				return fmt.Errorf("Plugin '%s' is invalid. Containers list 'spec.containers' must not contain more than 1 container, but '%d' found", meta.ID, len(meta.Spec.Containers))
-			}
-			if len(meta.Spec.Endpoints) != 0 {
-				return fmt.Errorf("Plugin '%s' is invalid. Setting endpoints at 'spec.endpoints' is not allowed in plugins of type '%s'", meta.ID, meta.Type)
-			}
+// GetPluginMeta downloads the metadata for a plugin. If specified,
+// defaultRegistry is used as the registry when plugin does not specify its registry.
+// If defaultRegistry is empty, and plugin does not specify a registry, an error is returned.
+func (b *Broker) GetPluginMeta(plugin model.PluginFQN, defaultRegistry string) (*model.PluginMeta, error) {
```
amisevsk: Similarly to `ValidateMeta` above, is there a reason to export this function but keep the other internal? To me it makes more sense to export `GetPluginMetas` and keep `getPluginMeta` internal.

brokers/unified/broker.go
davidfestal: Well, this is just the same error message as was before my change. I just moved it during the `Extract Function` refactoring.

brokers/unified/broker.go
davidfestal: An external program that would use brokers as a library could have the need to use this function on either 1 or several plugins.
Which is why the most useful public API call seemed to be the individual one. I didn't see any added value in exposing a function that simply loops on another basic API call.

brokers/unified/broker.go
davidfestal: An external program that would use brokers as a library could have the need to use this function on either 1 or several plugins.
Which is why the most useful public API call seemed to be the individual one. I didn't see any added value in exposing a function that simply loops on another basic API call.

brokers/unified/broker.go
davidfestal: Fixed in commit https://github.com/eclipse/che-plugin-broker/pull/62/commits/8587d25e8ab82466cb1ca0aaff5c0a083efa882f

brokers/unified/broker.go
amisevsk: It's strange from a design point of view to have a more general internal function that calls an exported specialized function, and is contrary to what we do in general (e.g. in upstream Che).

Won't every app using this as a library have to wrap the calls to `ValidateMeta` in a loop? Most libraries would work with a slice of `PluginMeta`, as you can always pass in a slice of one element.

brokers/unified/broker.go
amisevsk: If it's in the diff already, it may as well be improved.

davidfestal: Sure, but then shouldn't I change it in all error messages then ? Shouldn't it be preferably done in a distinct PR ?

brokers/unified/broker.go
davidfestal: The only project I know that uses the broker as a library is the Workspace CRD Controller :-) , and in this case, I manage each plugin separately because the overall plugin loop also includes other actions, such as creating the related k8s objects, etc... In fact it's simply the loop on devfile components. Sure I could refactor the code, but it's an example of the need to manage each plugin separately.
Additionally, to me proposing an atomic action in an API seems more general, or generic (in terms of use cases) than proposing to loop several times on the same action with a list of parameters, but here that's certainly a question of POV and wording ;-)

PR #69, opened 2019-08-13
Format code.

AndrienkoAleksandr:
### What does this PR do?
Format code 

Align code format for https://github.com/eclipse/che/issues/14126

### DIFF
--- a/common/connect.go
+++ b/common/connect.go
@@ -13,15 +13,15 @@
 package common
 
 import (
-	"net/http"
 	"log"
+	"net/http"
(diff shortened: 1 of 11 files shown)


common/connect.go
```diff
@@ -13,15 +13,15 @@
 package common
 
 import (
-	"net/http"
 	"log"
+	"net/http"
```
nickboldt: why isn't this list in alphabetical order like the list below? why are there two separate lists?

PR #70, opened 2019-08-21
Support relative paths

amisevsk:
### What does this PR do?
Adds support for relative paths in plugin meta.yaml extensions. This allows `meta.yaml`s in the plugin registry to specify e.g. 
```yaml
spec:
  extensions:
    - relative:extension/path/to/artifact
```
which will be resolved to `${PLUGIN_REGISTRY_URL}/path/to/artifact` during plugin broker execution.

I chose the format `relative:extension/` to mirror the vscode marketplace support already in place, with `vscode:extension/` although I'm open to suggestions.

### Testing this PR
A dockerimage built from this PR is available at `amisevsk/che-unified-plugin-broker:relative-paths`. I've also deployed a patched plugin registry with most `.vsix` and `.theia` artifacts cached and metas updated appropriately, available as `amisevsk/che-plugin-registry:cached-artifacts`.

```
CHE_WORKSPACE_PLUGIN__BROKER_UNIFIED_IMAGE=amisevsk/che-unified-plugin-broker:relative-paths
CHE_WORKSPACE_PLUGIN__REGISTRY__URL=http://che-plugin-registry-eclipse-che.devtools-dev.ext.devshift.net/v3
```

https://github.com/eclipse/che/issues/14163

### DIFF
--- a/brokers/unified/broker.go
+++ b/brokers/unified/broker.go
@@ -18,6 +18,8 @@ import (
 	"fmt"
 	"log"
 	"net/http"
+	"net/url"
+	"path"
 	"strings"
 
 	jsonrpc "github.com/eclipse/che-go-jsonrpc"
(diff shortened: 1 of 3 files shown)


amisevsk:
Additional info: the caching registry branch is here: https://github.com/amisevsk/che-plugin-registry/commits/cache-artifacts

nickboldt:
Dumb question...

given the devfile registry can say "gimme a plugin from /redhat/java/latest" with no protocol prefix, why can't we do the same here?

Why do we need "relative:extension/" ? could we not just assume that if the item starts with `/` it's local, but if it starts with `http` it's ... not local?

amisevsk:
@nickboldt It certainly could be done, but for this PR I tried to keep it as an explicit option. Reading a leading `/` to signify local seems a little opaque for me, since it triggers different behaviour from the plugin broker.

PR #73, opened 2019-09-08
Add ability override container entrypoint.

AndrienkoAleksandr:
### What does this PR do?
Add ability override container entrypoint.


model/model.go
```diff
@@ -99,6 +99,10 @@ type Container struct {
 	Ports        []ExposedPort `json:"ports" yaml:"ports"`
 	MemoryLimit  string        `json:"memoryLimit,omitempty" yaml:"memoryLimit,omitempty"`
 	MountSources bool          `json:"mountSources" yaml:"mountSources"`
+	// Base root command inside container
+	Command []string `json:"command" yaml:"command"`
```
amisevsk: Please line up `Args` and `Command` with other fields in the struct (I also think the comments are unnecessary)

PR #75, opened 2019-09-29
Start remote binary with override entry point

AndrienkoAleksandr:
### What does this PR do?
At this pr I changed a  bit model: applied Type field to the model.ChePlugin

Inject remote endpoint binary with override entrypoint sidecar containers. For this purpose we are using init container from che-theia editor-plugin:
```yaml
initContainers:
  - name: remote-runtime-injector
    image: aandrienko/che-theia-endpoint-runtime-binary:next
    volumes:
      - mountPath: "/remote-endpoint"
        name: remote-endpoint
        ephemeral: true
    env:
      - name: PLUGIN_REMOTE_ENDPOINT_EXECUTABLE
        value: /remote-endpoint/plugin-remote-endpoint
      - name: REMOTE_ENDPOINT_VOLUME_NAME
        value: remote-endpoint
```

Plugin broker should find this init container by name `remote-runtime-injector` and only for che-theia editor plugin. Then plugin-broker should analize env variables and volume from init container.

`REMOTE_ENDPOINT_VOLUME_NAME` env we are using to find one volume which should be shared with other sidecar containers. In this volume init container on start should copy remote binary

`PLUGIN_REMOTE_ENDPOINT_EXECUTABLE` env will be shared with all sidecar containers.

Also che-plugin-broker override entypoint for all sidecar containers to start remote plugin. **But if sidecar container has in the meta.yaml already defined entrypoint**  che-plugin broker don't override it. It's mean that plugin writer could define do something and then start remote plugin(using env `PLUGIN_REMOTE_ENDPOINT_EXECUTABLE`). 

### DIFF
--- a/brokers/unified/remote_runtime_injector.go
+++ b/brokers/unified/remote_runtime_injector.go
@@ -20,12 +20,12 @@ import (
 )
 
 const (
-	DefaultEditorName = "che-theia"
+	CheTheiaEditorName = "che-theia"
 
 	InjectorContainerName = "remote-runtime-injector"
 
-	RemoteEndPontExecutableEnvVar = "PLUGIN_REMOTE_ENDPOINT_EXECUTABLE"
-	VolumeNameEnvVar              = "REMOTE_ENDPOINT_VOLUME_NAME"
+	RemoteEndPointExecutableEnvVar = "PLUGIN_REMOTE_ENDPOINT_EXECUTABLE"
+	VolumeNameEnvVar               = "REMOTE_ENDPOINT_VOLUME_NAME"
 )
 
 type RemotePluginInjection struct {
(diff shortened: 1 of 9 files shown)


AndrienkoAleksandr:
> Looking at this PR vs #76, this one should be closed, right?

Both PRs #76 and #75 works fine, but yes one of them should be closed after discussion.

AndrienkoAleksandr:
Description updated

AndrienkoAleksandr:
@amisevsk @davidfestal @sleshchenko @l0rd pr update and ready for review.

brokers/unified/remote_runtime_injector.go
```diff
@@ -0,0 +1,147 @@
...
+
+	"github.com/eclipse/che-plugin-broker/model"
+)
+
+const (
+	CheTheiaEditorName = "che-theia"
+
+	InjectorContainerName = "remote-runtime-injector"
+
+	RemoteEndPointExecutableEnvVar = "PLUGIN_REMOTE_ENDPOINT_EXECUTABLE"
```
sleshchenko: `EndPoint` -> `Endpoint`

PR #76, opened 2019-09-30
Start remote binary using env and appling ephemeral volume

AndrienkoAleksandr:
### What does this PR do?
At this pr I changed a  bit model: applied Type field to the model.ChePlugin

Inject remote endpoint binary using env variable with remote binary path and volume where is binary  located.
For this purpose we are using init container from che-theia editor-plugin:

```yaml
initContainers:
  - name: remote-runtime-injector
    image: aandrienko/che-theia-endpoint-runtime-binary:next
    volumes:
      - mountPath: "/remote-endpoint"
        name: remote-endpoint
        ephemeral: true
    env:
      - name: PLUGIN_REMOTE_ENDPOINT_EXECUTABLE
        value: /remote-endpoint/plugin-remote-endpoint
      - name: REMOTE_ENDPOINT_VOLUME_NAME
        value: remote-endpoint
```

Plugin broker should find this init container by name `remote-runtime-injector` and only for che-theia editor plugin. Then plugin-broker should analize env variables and volume from init container.

`REMOTE_ENDPOINT_VOLUME_NAME` env we are using to find one volume which should be shared with other sidecar containers. In this volume init container on start should copy remote binary

`PLUGIN_REMOTE_ENDPOINT_EXECUTABLE` env will be shared with all sidecar containers.

To start remote binary in the sidecar container, plugin writer should define in the sidecar dockerfile ENTRYPOINT:

```Dockerfile
FROM some-image 
....
ENTRYPOINT ${PLUGIN_REMOTE_ENDPOINT_EXECUTABLE} 
```

Or apply starting remote binary in the entrypoint.sh script, if plugin image uses such one.

### DIFF
--- a/brokers/unified/remote_runtime_injector.go
+++ b/brokers/unified/remote_runtime_injector.go
@@ -20,12 +20,12 @@ import (
 )
 
 const (
-	DefaultEditorName = "che-theia"
+	CheTheiaEditorName = "che-theia"
 
 	InjectorContainerName = "remote-runtime-injector"
 
-	RemoteEndPontExecutableEnvVar = "PLUGIN_REMOTE_ENDPOINT_EXECUTABLE"
-	VolumeNameEnvVar              = "REMOTE_ENDPOINT_VOLUME_NAME"
+	RemoteEndPointExecutableEnvVar = "PLUGIN_REMOTE_ENDPOINT_EXECUTABLE"
+	VolumeNameEnvVar               = "REMOTE_ENDPOINT_VOLUME_NAME"
 )
 
 type RemotePluginInjection struct {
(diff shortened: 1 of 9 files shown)


benoitf:
IMHO description of the PR should tell the name of the env variable, what to add, etc.
Why it's like that, what we can do, with an example, etc.

nickboldt:
What is the impact on docs for this change?

brokers/unified/remote_runtime_injector.go
```diff
@@ -0,0 +1,131 @@
...
+
+	"github.com/eclipse/che-plugin-broker/model"
+)
+
+const (
+	DefaultEditorName = "che-theia"
+
+	InjectorContainerName = "remote-runtime-injector"
+
+	RemoteEndPontExecutableEnvVar = "PLUGIN_REMOTE_ENDPOINT_EXECUTABLE"
```
amisevsk: Typo: RemoteEndPo**i**ntExecutableEnvVar

brokers/unified/remote_runtime_injector.go
```diff
@@ -0,0 +1,131 @@
...
+	Volume  model.Volume
+	Env     model.EnvVar
+	Command []string
+	Args    []string
+}
+
+func InjectRemoteRuntime(plugins []model.ChePlugin) {
+	editorPlugin, err := findDefaultEditorPlugin(plugins)
+	if err != nil {
+		return
```
amisevsk: We shouldn't fail silently here.

brokers/unified/remote_runtime_injector.go
```diff
@@ -0,0 +1,131 @@
...
+
+func getRuntimeInjection(editorPlugin *model.ChePlugin) (*RemotePluginInjection, error) {
+	containerInjector, err := findContainerInjector(editorPlugin.InitContainers)
+	if err != nil {
+		return nil, err
+	}
+
+	runtimeBinaryPathEnv, err := findEnv(RemoteEndPontExecutableEnvVar, containerInjector.Env)
+	if err != nil || runtimeBinaryPathEnv.Value == "" {
+		return nil, err
```
amisevsk: If `err == nil` and `runtimeBinaryPathEnv.Value == ""`, this will return `nil, nil`.

AndrienkoAleksandr:
> IMHO description of the PR should tell the name of the env variable, what to add, etc.
Why it's like that, what we can do, with an example, etc.

Fixed

brokers/unified/remote_runtime_injector.go
AndrienkoAleksandr: Fixed

AndrienkoAleksandr:
Closed on favor full solution https://github.com/eclipse/che-plugin-broker/pull/75
@amisevsk all your comments I handled using cherry-pick commits to https://github.com/eclipse/che-plugin-broker/pull/75

PR #78, opened 2019-10-21
add rhel.Dockerfiles as copied from...

nickboldt:
add rhel.Dockerfiles as copied from downstream

Change-Id: I12600bdac3d033212bfa58aa5deb629dceefdada
Signed-off-by: nickboldt <nboldt@redhat.com>

### DIFF
--- a/build/unified/rhel.Dockerfile
+++ b/build/unified/rhel.Dockerfile
@@ -1,0 +1,30 @@
+# Copyright (c) 2019 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#   Red Hat, Inc. - initial API and implementation
+#
+
+# https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/rhel8/go-toolset
+FROM registry.access.redhat.com/rhel8/go-toolset:1.11.13-18 as builder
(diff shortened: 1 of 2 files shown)


nickboldt:
We *could* remove it but a) it doesn't hurt anyone and b) makes integration/sync with downstream easier. 

Don't fear my harmless metadata, dude. :)

nickboldt:
I've removed the metadata.

build/unified/rhel.Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#   Red Hat, Inc. - initial API and implementation
+#
+
+# https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/rhel8/go-toolset
+FROM registry.access.redhat.com/rhel8/go-toolset:1.11.13-18 as builder
```
benoitf: $ docker pull registry.access.redhat.com/rhel8/go-toolset:1.11.13-18
Error response from daemon: unsupported: This repo requires terms acceptance and is only available on registry.redhat.io

AFAIK it can't be upstream then

nickboldt: well, we can switch it to reg.rh.io ... or go back to using rhel7 go toolset to build. which would you prefer?

nickboldt:
I've switched to use rhel7 go-toolset as it's on reg.acc.rh.com and requires no authentication to pull images. So it's DIFFERENT from what I'm using in CRW (yay, forks!) but it's close enough so at least there's a rhel version of the plugin brokers in Che now. \o/ can I get a +1?

PR #81, opened 2019-12-23
Add master, nightly and release scripts for centOS CI jobs

sleshchenko:
### What does this PR do?
This PR removes scripts that were used by CodenvyCI and adds master, nightly and release scripts for centOS CI jobs.
The corresponding jobs are added in https://github.com/openshiftio/openshiftio-cico-jobs/pull/1105
https://github.com/eclipse/che/issues/14378


VERSION
```diff
@@ -0,0 +1 @@
+3.1.0-SNAPHOT
```
ibuziuk: has `3.0.0` version been released already?

sleshchenko: Not yet. I'm going to do it as soon as webhooks will be set.
Do you think it makes sense to leave 3.0.0-SNAPSHOT now?

ibuziuk: yeah, IMO ` 3.0.0-SNAPSHOT` is a better choice

sleshchenko:
I'm merging this PR since Draft with jobs is already merged. But feel free to comment and I'll address it in upcoming  PRs.

PR #86, opened 2020-01-09
Improve how already-downloaded plugin extensions are handled.

amisevsk:
### What does this PR do?
Allows caching of installed plugin extensions in the workspace. Instead of erasing the entire `/plugins` directory on every start, the broker will instead store a json log of installed plugins, and use this to only download required extensions, via the rule:

1. If a plugin is installed locally but not in the current workspace, all of its artifacts are removed
2. If a plugin is not installed locally and included in the current workspace, all of its artifacts will be downloaded
3. If a plugin is installed but the artifacts to be downloaded are different compared to a prior run (e.g. a new version of `latest`), then 
    - Extensions that are no longer needed are removed
    - New extensions are downloaded
    - Extensions that are shared between the old and new version are reused.

These changes make the assumption that all extension artifacts are served from unique URLs, and will fail if e.g. the `.vsix` file at a URL is updated.

Installed plugin metadata is stored in a new struct (see `model.InstalledPlugin`), which stores the plugin ID, whether the plugin is remote (runs in a sidecar), and a map of extension URLs to filesystem paths. A list of installed plugins is saved in the `/plugins` directory along with an internal file version number to allow versioned changes.

This PR significantly improves brokering time for the most common use case in starting a workspace, where a previously started non-ephemeral workspace is started a second time; in this case brokering is complete almost immediately, since nothing needs to be downloaded.

https://github.com/eclipse/che/issues/15470

### DIFF
--- a/brokers/artifacts/broker.go
+++ b/brokers/artifacts/broker.go
@@ -14,15 +14,12 @@ package artifacts
 
 import (
 	"fmt"
-	"path/filepath"
 
 	jsonrpc "github.com/eclipse/che-go-jsonrpc"
 	"github.com/eclipse/che-plugin-broker/common"
 	"github.com/eclipse/che-plugin-broker/model"
 	"github.com/eclipse/che-plugin-broker/utils"
 )
-
-const errorNoExtFieldsTemplate = "Field 'extensions' is not found in the description of the plugin '%s'"
 
 // Broker is used to process Che plugins
 type Broker struct {
@@ -58,122 +55,55 @@ func (b *Broker) Start(pluginFQNs []model.PluginFQN, defaultRegistry string) err
 	b.PubStarted()
 	b.PrintInfo("Starting plugin artifacts broker")
 
-	b.cleanupPluginsDirectory()
 	pluginMetas, err := utils.GetPluginMetas(pluginFQNs, defaultRegistry, b.ioUtils)
 	if err != nil {
 		return b.fail(fmt.Errorf("Failed to download plugin meta: %s", err))
 	}
-	b.PrintInfo("Downloading plugin extensions")
 
 	err = utils.ResolveRelativeExtensionPaths(pluginMetas, defaultRegistry)
 	if err != nil {
 		return b.fail(err)
 	}
 
-	for _, meta := range pluginMetas {
-		err = b.ProcessPlugin(meta)
+	requestedPlugins := convertMetasToPlugins(pluginMetas)
+
+	toInstall := b.syncWithPluginsDir(requestedPlugins)
+
+	for _, plugin := range toInstall {
+		err = b.ProcessPlugin(&plugin)
 		if err != nil {
 			return b.fail(err)
 		}
 	}
 
+	err = b.writeInstalledPlugins(toInstall)
+	if err != nil {
+		b.PrintInfo("WARN: Failed to log installed plugins: %s", err)
+	}
+
 	b.PrintInfo("All plugin artifacts have been successfully downloaded")
 	b.PubDone("")
-
 	return nil
 }
 
-func (b *Broker) cleanupPluginsDirectory() {
-	b.PrintInfo("Cleaning /plugins dir")
-	files, err := b.ioUtils.GetFilesByGlob(filepath.Join("/plugins", "*"))
-	if err != nil {
-		// Send log about clearing failure but proceed.
-		// We might want to change this behavior later
-		b.PrintInfo("WARN: failed to clear /plugins directory. Error: %s", err)
-		return
-	}
+func convertMetasToPlugins(metas []model.PluginMeta) []model.CachedPlugin {
+	plugins := make([]model.CachedPlugin, 0)
 
-	for _, file := range files {
-		err = b.ioUtils.RemoveAll(file)
-		if err != nil {
-			b.PrintInfo("WARN: failed to remove '%s'. Error: %s", file, err)
+	for _, meta := range metas {
+		if !utils.IsTheiaOrVscodePlugin(meta) {
+			continue
 		}
-	}
-}
-
-// ProcessPlugin processes metas of different plugin types and passes metas of each particular type
-// to the appropriate plugin broker
-func (b *Broker) ProcessPlugin(meta model.PluginMeta) error {
-	if !utils.IsTheiaOrVscodePlugin(meta) {
-		return nil
-	}
-
-	URLs, err := getUrls(meta)
-	if err != nil {
-		return err
-	}
-
-	workDir, err := b.ioUtils.TempDir("", "vscode-extension-broker")
-	if err != nil {
-		return err
-	}
-
-	archivesPaths, err := b.downloadArchives(URLs, meta, workDir)
-	if err != nil {
-		return err
-	}
-
-	err = b.injectPlugin(meta, archivesPaths)
-	return err
-}
-
-func (b *Broker) downloadArchives(URLs []string, meta model.PluginMeta, workDir string) ([]string, error) {
-	paths := make([]string, 0)
-	for i, URL := range URLs {
-		archivePath := b.ioUtils.ResolveDestPathFromURL(URL, workDir)
-		b.PrintDebug("Downloading VS Code extension archive '%s' for plugin '%s' to '%s'", URL, meta.ID, archivePath)
-		b.PrintInfo("Downloading VS Code extension %d/%d for plugin '%s'", i+1, len(URLs), meta.ID)
-		archivePath, err := b.ioUtils.Download(URL, archivePath, true)
-		paths = append(paths, archivePath)
-		if err != nil {
-			return nil, fmt.Errorf("failed to download plugin from %s: %s", URL, err)
-		}
-	}
-	return paths, nil
-}
-
-func (b *Broker) injectPlugin(meta model.PluginMeta, archivesPaths []string) error {
-	for _, path := range archivesPaths {
-		pluginPath := "/plugins"
+		plugin := model.CachedPlugin{}
+		plugin.ID = meta.ID
 		if len(meta.Spec.Containers) > 0 {
-			// Plugin is remote
-			pluginUniqueName := utils.GetPluginUniqueName(meta)
-			pluginPath = filepath.Join(pluginPath, "sidecars", pluginUniqueName)
-			err := b.ioUtils.MkDir(pluginPath)
-			if err != nil {
-				return err
-			}
+			plugin.IsRemote = true
 		}
-		pluginArchiveName := b.generatePluginArchiveName(meta, path)
-		pluginArchivePath := filepath.Join(pluginPath, pluginArchiveName)
-		err := b.ioUtils.CopyFile(path, pluginArchivePath)
-		if err != nil {
-			return err
+		plugin.CachedExtensions = make(map[string]string)
+		for _, ext := range meta.Spec.Extensions {
+			plugin.CachedExtensions[ext] = ""
 		}
+		plugins = append(plugins, plugin)
 	}
-	return nil
-}
 
-func (b *Broker) generatePluginArchiveName(plugin model.PluginMeta, archivePath string) string {
-	archiveName := filepath.Base(archivePath)
-	return fmt.Sprintf("%s.%s.%s.%s.%s", plugin.Publisher, plugin.Name, plugin.Version, b.rand.String(10), archiveName)
-}
-
-func getUrls(meta model.PluginMeta) ([]string, error) {
-	URLs := make([]string, 0)
-	if len(meta.Spec.Extensions) == 0 {
-		return nil, fmt.Errorf(errorNoExtFieldsTemplate, meta.ID)
-	}
-	URLs = append(URLs, meta.Spec.Extensions...)
-	return URLs, nil
+	return plugins
 }
--- a/brokers/artifacts/extensions.go
+++ b/brokers/artifacts/extensions.go
@@ -1,0 +1,104 @@
+//
+// Copyright (c) 2019 Red Hat, Inc.
+// This program and the accompanying materials are made
+// available under the terms of the Eclipse Public License 2.0
+// which is available at https://www.eclipse.org/legal/epl-2.0/
+//
+// SPDX-License-Identifier: EPL-2.0
+//
+// Contributors:
+//   Red Hat, Inc. - initial API and implementation
+//
+
+package artifacts
+
+import (
+	"fmt"
+	"path/filepath"
+	"strings"
+
+	"github.com/eclipse/che-plugin-broker/model"
+	"github.com/eclipse/che-plugin-broker/utils"
+)
+
+// ProcessPlugin downloads all undownloaded plugin extensions and places the
+// relevant artifacts in their appropriate location in the /plugins directory.
+// If a plugin already has artifacts downloaded for a given extension, that
+// extension is skipped.
+func (b *Broker) ProcessPlugin(plugin *model.CachedPlugin) error {
+	workDir, err := b.ioUtils.TempDir("", "artifacts-broker")
+	if err != nil {
+		return err
+	}
+
+	// Workaround: messages can be displayed out of order in the workspace loading page
+	// Collect messages in a buffer and print them as a single string when needed.
+	logBuf := make([]string, 0)
+	defer b.flushLog(&logBuf)
+
+	logBuf = append(logBuf, fmt.Sprintf("Processing plugin %s", plugin.ID))
+	numExtensions := len(plugin.CachedExtensions)
+	extensionIdx := 0
+	for URL, path := range plugin.CachedExtensions {
+		extensionIdx = extensionIdx + 1
+		logBuf = append(logBuf, fmt.Sprintf("  Installing plugin extension %d/%d", extensionIdx, numExtensions))
+		if path != "" {
+			logBuf = append(logBuf, fmt.Sprintf("    Plugin already downloaded"))
+			continue
+		}
+		logBuf = append(logBuf, fmt.Sprintf("    Downloading plugin from %s", URL))
+		logBuf = b.flushLog(&logBuf)
+		archivePath, err := b.downloadArchive(URL, plugin.ID, workDir)
+		if err != nil {
+			return err
+		}
+		pluginPath, err := b.injectPlugin(plugin, archivePath)
+		if err != nil {
+			return err
+		}
+		plugin.CachedExtensions[URL] = pluginPath
+	}
+	return nil
+}
+
+func (b *Broker) downloadArchive(URL string, pluginID string, workDir string) (string, error) {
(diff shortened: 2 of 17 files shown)


brokers/artifacts/extensions.go
```diff
@@ -0,0 +1,148 @@
...
+// plugins to be installed. Any plugins that are currently installed but not requested in the workspace will
+// be removed from the filesystem.
+func (b *Broker) getPlan(requested, installed []model.CachedPlugin) (toInstall []model.CachedPlugin) {
+	toInstall = requested
+	for _, plugin := range installed {
+		match := findPlugin(plugin, requested)
+		if match == nil {
+			// Plugin has been uninstalled since last start
+			b.PrintInfo("Uninstalling plugin: %s", plugin.ID)
+			if err := b.removePlugin(plugin); err != nil {
```
sleshchenko: It was quite unexpected for me that `getPlan` method actually removes plugins, maybe instead of performing any operation here it makes sense to keep it only analyzing extenstions and return plan:
```
func (b *Broker) getPlan(requested, installed []model.CachedPlugin) (toInstall []model.CachedPlugin, toRemove []model.CachedPlugin, filesToRemove []File) {
```
and then remove them on another level? Or maybe there can be found a more clear method name.

P.S. I have not had reviewed your changes deeply enough if my comment does not make any sense - please just mark it resolved =)

amisevsk: You're right that the name is not very good, I will improve that. The issue I ran into with trying to get a plan and then execute it is that there are three cases:
1. A plugin is added to the workspace -> download its extensions
2. A plugin is removed from the workspace -> remove its extensions
3. A plugin is updated in the workspace -> remove outdated extensions, download new ones

The third case is hard to support and also potentially the most common (`latest` represents an "update", sort of). What I opted for, as a result, is to partially fill plugins in the plan to support case three (if an extension has already been downloaded, it has the extension's path added to the struct).

Looking back this is probably over-engineered, since it only occurs when e.g. jdt is updated but the debug extension is not.

brokers/artifacts/broker.go
```diff
@@ -58,122 +57,125 @@ func (b *Broker) Start(pluginFQNs []model.PluginFQN, defaultRegistry string) err
...
 	err = utils.ResolveRelativeExtensionPaths(pluginMetas, defaultRegistry)
 	if err != nil {
 		return b.fail(err)
 	}
 
-	for _, meta := range pluginMetas {
-		err = b.ProcessPlugin(meta)
+	requestedPlugins := convertMetasToPlugins(pluginMetas)
+
+	toInstall := b.SyncPlugins(requestedPlugins)
```
sleshchenko: It makes an impression that not every plugin that is requested should be installed.
But later I found 
```go
func (b *Broker) getPlan(requested, installed []model.CachedPlugin) (toInstall []model.CachedPlugin) {
	toInstall = requested
        ...
```
Maybe then `b.SyncPlugins(requestedPlugins)` should return nothing...
Or consumes Metas and returns `CachedPlugin` ?

brokers/artifacts/broker.go
```diff
@@ -58,122 +57,125 @@ func (b *Broker) Start(pluginFQNs []model.PluginFQN, defaultRegistry string) err
...
 
-func (b *Broker) cleanupPluginsDirectory() {
-	b.PrintInfo("Cleaning /plugins dir")
-	files, err := b.ioUtils.GetFilesByGlob(filepath.Join("/plugins", "*"))
+// ProcessPlugin downloads all undownloaded plugin extensions and places the
+// relevant artifacts in their appropriate location in the /plugins directory.
+// If a plugin already has artifacts downloaded for a given extension, that
+// extension is skipped.
+func (b *Broker) ProcessPlugin(plugin *model.CachedPlugin) error {
+	workDir, err := b.ioUtils.TempDir("", "vscode-extension-broker")
```
sleshchenko: Just wondering if it would be better to rename the folder to `artifacts-broker` since there is no explicit VSCode related logic in code, everything in name in a general manner - Plugin, Extension.

brokers/artifacts/broker.go
```diff
@@ -58,122 +57,125 @@ func (b *Broker) Start(pluginFQNs []model.PluginFQN, defaultRegistry string) err
...
-	for _, file := range files {
-		err = b.ioUtils.RemoveAll(file)
+	b.PrintInfo("Processing plugin %s", plugin.ID)
+	numExtensions := len(plugin.CachedExtensions)
+	extensionIdx := 0
+	for URL, path := range plugin.CachedExtensions {
+		extensionIdx = extensionIdx + 1
+		b.PrintInfo("  Installing plugin extension %d/%d", extensionIdx, numExtensions)
+		if path != "" {
+			// Plugin has previously been downloaded, nothing needs to be done.
```
sleshchenko: maybe it makes sense to print log that already downloaded archive is used

brokers/artifacts/broker.go
```diff
@@ -58,122 +57,125 @@ func (b *Broker) Start(pluginFQNs []model.PluginFQN, defaultRegistry string) err
...
 }
 
-func (b *Broker) cleanupPluginsDirectory() {
-	b.PrintInfo("Cleaning /plugins dir")
-	files, err := b.ioUtils.GetFilesByGlob(filepath.Join("/plugins", "*"))
+// ProcessPlugin downloads all undownloaded plugin extensions and places the
+// relevant artifacts in their appropriate location in the /plugins directory.
+// If a plugin already has artifacts downloaded for a given extension, that
+// extension is skipped.
+func (b *Broker) ProcessPlugin(plugin *model.CachedPlugin) error {
```
sleshchenko: AFAIU previously this method also generated configuration for Theia and VsCode plugins, but now it does not. Maybe it makes sense to rename this method to DownloadExtenstions?

brokers/artifacts/broker.go
```diff
@@ -58,122 +57,125 @@ func (b *Broker) Start(pluginFQNs []model.PluginFQN, defaultRegistry string) err
...
 
-func (b *Broker) injectPlugin(meta model.PluginMeta, archivesPaths []string) error {
-	for _, path := range archivesPaths {
-		pluginPath := "/plugins"
+func convertMetasToPlugins(metas []model.PluginMeta) []model.CachedPlugin {
+	plugins := make([]model.CachedPlugin, 0)
+
+	for _, meta := range metas {
+		if !utils.IsTheiaOrVscodePlugin(meta) {
+			continue
```
sleshchenko: In terms of Che Plugins model, even non-Theia or non-Vscode is a Che Plugin. Please consider putting here a comment why only Theia or VsCode plugins are returned here - like it's artifacts broker and only they have such...

brokers/artifacts/broker.go
amisevsk: :+1: this is a holdover from before the refactor, will update.

amisevsk: :+1:

sleshchenko: > :+1: 

:+1: 

;)

brokers/artifacts/broker.go
amisevsk: The idea behind `SyncPlugins` is that it syncs the list of requested plugins with what's on disk, i.e.
- Requested plugins should contain already-installed extensions
- Disk should not contain plugins that are not requested

I think the unclear thing is that the convention for `model.CachedPlugin` is that if a extension already has a path, it doesn't need to be downloaded. We set `toInstall = requested` since at the end of the day, all requested plugins must be installed, and then we partially fill `toInstall`'s extension urls with already downloaded paths.

If we make `b.SyncPlugins(requestedPlugins)` return nothing, then we basically move all the functionality in `broker.go` into `extensions.go`. Currently the division of labour is 
- extensions.go: keep up local disk cache of plugins
- broker.go: download extensions when necessary

As an example, suppose a workspace has `redhat/vscode-openshift-connector/latest` installed, and a Che update has caused `latest` to go from `0.0.21` to `0.1.0`. Running `SyncPlugins` will return
```go
model.CachedPlugin{
  ID: redhat/vscode-openshift-connector/latest
  IsRemote: true
  CachedExtensions: map[string]string {
    "https://github.com/[...]/vscode-kubernetes-tools-1.0.0.vsix": "<path to downloaded vsix>"
    "https://github.com/[...]/redhat.vscode-yaml-0.4.0.vsix": "<path to downlaoded vsix>"
    "https://github.com/[...]/openshift-connector-0.0.21-228.vsix": "" // empty string means should be downloaded
  }
}
```

Then `ProcessPlugin` will only download `https://github.com/[...]/openshift-connector-0.0.21-228.vsix` and update that entry in the map.

brokers/artifacts/broker.go
amisevsk: :+1:

brokers/artifacts/broker.go
amisevsk: Done!

brokers/artifacts/broker.go
amisevsk: Renamed method to `SyncWithPluginsDir`, hopefully that makes things a little clearer.

brokers/artifacts/extensions.go
nickboldt: What about scenarios where the same extension (but different versions) are installed because of plugins which depend on different versions? eg., 
```
redhat/java8/latest 
- https://download.jboss.org/jbosstools/vscode/3rdparty/vscode-java-debug/vscode-java-debug-0.20.0.vsix

redhat/quarkus-java8
- https://download.jboss.org/jbosstools/vscode/3rdparty/vscode-java-debug/vscode-java-debug-0.23.0.vsix
```

Do you make sure that an update for one plugin (eg., newer quarkus results in 0.23.0 -> 0.24.0) doesn't also garbage collect the 0.2.0 used by the other plugin?

amisevsk: Currently, each plugin gets its own .vsix on its own path, so there's no overlap for that case. E.g. redhat/java plugins end up in
```
/plugins
└── sidecars
    └── redhat_java_0_43_0
        ├── redhat.java.0.43.0.kmwjfvvtpc.java-0.43.0-1473.vsix
        └── redhat.java.0.43.0.wgfmdnrsmo.vscode-java-debug-0.16.0.vsix
```

brokers/artifacts/extensions.go
```diff
@@ -0,0 +1,148 @@
...
+		return nil, err
+	}
+	if pluginsJSON.Version != installedPluginsJSONVersion {
+		b.PrintInfo("Installed plugins cache is incompatible with current version of plugin broker. All plugins will be redownloaded.")
+		return nil, fmt.Errorf("Installed plugins list is from previous version of broker")
+	}
+	return pluginsJSON.Plugins, nil
+}
+
+func (b *Broker) writeInstalledPlugins(cached []model.CachedPlugin) error {
```
sleshchenko: It's used outside of this file, maybe then it should be public as `SyncWithPluginsDir` is?

brokers/artifacts/broker.go
sleshchenko: > If we make b.SyncPlugins(requestedPlugins) return nothing, then we basically move all the functionality in broker.go into extensions.go. Currently the division of labour is
> 
> extensions.go: keep up local disk cache of plugins
> broker.go: download extensions when necessary

Just an idea to consider:

* broker.go: publish events, control downloading extensions and caching process by invoking the corresponding methods from files below
* extensions.go: download extensions when necessary
* cache.go: keep up local disk cache of plugins

Just minor comment what may be considered to be implemented in upcoming PRs.

amisevsk: That makes a lot of sense, thanks.

amisevsk:
@sleshchenko I've updated the PR (but forgot to split it into separate commits so it's hard to read -- sorry). Changes:
- Check if the error in reading installed_plugins.json is a "file does not exist error", and don't print the error message to the user in that case (https://github.com/eclipse/che-plugin-broker/pull/86#pullrequestreview-342463933)
- Reorganize the broker to split into files as you suggested (`broker.go`, `extensions.go`, and `cache.go`)
  - I reverted the name change `ProcessPlugin` -> `DownloadExtensions`, since that's technically a backwards-incompatible change from 3.0.0.
  - I also made the `cache.go` functions internal, since they really should be only used as part of the broker. The only exported functions in the broker are now `Start`, `PushEvents` and `ProcessPlugin`. None of the file syncing stuff is exported, since there's no way to easily expose a uniform interface -- one step has to happen before `DownloadExtensions` and the other has to happen after.

sleshchenko:
> I reverted the name change ProcessPlugin -> DownloadExtensions, since that's technically a backwards-incompatible change from 3.0.0.

Well, as far as I see `ProcessPlugin` comes from `common.BrokerImpl` that is not used anymore.
So, technically it's backward-incompatible but there are not much depending on Che Plugin Broker .
It seems a good time to review what we want to expose as public API of Plugin Broker... Is it `BrokerImpl`? If no - then I think we can afford such breaking change.
It definitely can be considered to implement later in the scope of the issue about making Plugin Broker usable as a library.

amisevsk:
In testing these changes on an actual cluster, I encountered a few issues that are fixed in the 3 new commits:
- 9d4f58b reworks how messages are pushed through the event bus, since logging multiple times quickly can cause messages to be out of order
- 5ca4b9c fixes a bug where removing a local plugin removed all remote plugins (due to erasing everything in `/plugins`)
- c4144e7 is a sanity check -- we ensure all cached extensions are present on disk before assuming we can use them

Scenarios tested:
- Ephemeral workspace (all plugins always downloaded)
- Non-ephemeral workspace
    - Add remote plugin to existing workspace (plugin is downloaded)
    - Add local plugin to existing workspace (plugin is downloaded)
    - Remove remote plugin (plugin extensions are removed, others untouched)
    - Remove local plugin (plugin extensions are removed, others untouched)
    - Delete `installed.json` -> `/plugins` erased and all plugins redownloaded
    - Delete an extension from an installed plugin -> `/plugins` erased and all plugins redownloaded.

There is still one outstanding issue, where the logs for `Saving log of installed plugins` is delivered before the final plugin is checked (I see it with java-maven + the vim extension).

@sleshchenko please take another look, if you have a moment.

amisevsk:
Opened issue https://github.com/eclipse/che/issues/15758 to address the out of order logs.

brokers/artifacts/extensions.go
```diff
@@ -0,0 +1,95 @@
...
+	numExtensions := len(plugin.CachedExtensions)
+	extensionIdx := 0
+	for URL, path := range plugin.CachedExtensions {
+		extensionIdx = extensionIdx + 1
+		infoLine.WriteString(fmt.Sprintf("\n  Installing plugin extension %d/%d", extensionIdx, numExtensions))
+		if path != "" {
+			infoLine.WriteString(fmt.Sprintf("\n    Plugin already downloaded"))
+			continue
+		}
+		infoLine.WriteString(fmt.Sprintf("\n    Downloading plugin from %s", URL))
```
sleshchenko: With it, we get ordered output but we see nothing when something is happening, like downloading...

Maybe it would be better if we flush buffer each time before the time-consuming task is started?

like:
```go
        var infoLine strings.Builder
	defer b.PrintInfoBuffer(&infoLine)

	infoLine.WriteString(fmt.Sprintf("Processing plugin %s", plugin.ID))
	numExtensions := len(plugin.CachedExtensions)
	extensionIdx := 0
	for URL, path := range plugin.CachedExtensions {
		extensionIdx = extensionIdx + 1
		infoLine.WriteString(fmt.Sprintf("\n  Installing plugin extension %d/%d", extensionIdx, numExtensions))
		if path != "" {
			infoLine.WriteString(fmt.Sprintf("\n    Plugin already downloaded"))
			continue
		}
		infoLine.WriteString(fmt.Sprintf("\n    Downloading plugin from %s", URL))

                b.PrintInfoBuffer(&infoLine)
                infoLine.reset()
                ...
```
WDYT?

sleshchenko: Another implementation I can imagine with channels, where one thread receives messages and collects in into buffer which is flushed like after some period, like 1-2 seconds.
But it would require even more changes in this PR, I'm OK to go with the current implementation.

brokers/artifacts/broker.go
```diff
@@ -58,122 +55,55 @@ func (b *Broker) Start(pluginFQNs []model.PluginFQN, defaultRegistry string) err
...
 
-	archivesPaths, err := b.downloadArchives(URLs, meta, workDir)
-	if err != nil {
-		return err
-	}
-
-	err = b.injectPlugin(meta, archivesPaths)
-	return err
+	b.PrintInfo("All plugin artifacts have been successfully downloaded")
+	b.PubDone("")
```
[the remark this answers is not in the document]
amisevsk: I think `b.PubDone()` doesn't display output, e.g. for the metadata broker it's used to pass the Che plugins json back to Che server. The empty `PubDone()` notifies the server that brokering completed successfully, but there is no tooling to return and Che server is not listening for it (I'm not certain Che server even cares if we send a success event, but it's probably good practice).

brokers/artifacts/extensions.go
amisevsk: I agree that not showing anything while downloading is a problem, this is mostly a workaround to preserve logs in case of failure. With a slower internet connection, it does look strange, though. 

I like your suggestion of attempting to dump info more often, since as it's currently implemented the java plugin logs *both* extensions at once, which isn't ideal. The main goal is to keep the `downloading plugin x/#` logs for each extension together to avoid interleaving.

I've pushed a commit to rework this to log on every loop iteration and also move to using `[]string` in order to avoid worrying about newlines.

brokers/artifacts/extensions.go
```diff
@@ -31,20 +31,24 @@ func (b *Broker) ProcessPlugin(plugin *model.CachedPlugin) error {
...
 	numExtensions := len(plugin.CachedExtensions)
 	extensionIdx := 0
 	for URL, path := range plugin.CachedExtensions {
 		extensionIdx = extensionIdx + 1
-		infoLine.WriteString(fmt.Sprintf("\n  Installing plugin extension %d/%d", extensionIdx, numExtensions))
+		logBuf = append(logBuf, fmt.Sprintf("  Installing plugin extension %d/%d", extensionIdx, numExtensions))
 		if path != "" {
-			infoLine.WriteString(fmt.Sprintf("\n    Plugin already downloaded"))
+			logBuf = append(logBuf, fmt.Sprintf("    Plugin already downloaded"))
+			b.flushLog(logBuf)
```
sleshchenko: When plugin is already downloaded iteration should be really fast, I guess we don't really need to flush buffer here.
But I'm OK with flushing it here if it does not lead to propagating logs in the wrong order.

amisevsk: You're right, it was actually a mistake in `b.flushLog` that caused me to put this line here (the deferred flush gets the just-initialized buffer so it always logged nothing). This is fixed now.

PR #88, opened 2020-01-22
Bump version to 3.2.0-SNAPSHOT

amisevsk:
### What does this PR do?
Bumps version to 3.2.0-SNAPSHOT, following the release of 3.1.0.


amisevsk:
@sleshchenko WDYT about using `nightly` as version instead of `x.y.z-SNAPSHOT` to avoid having to modify this file on every release?

sleshchenko:
On the one hand it allows us to avoid modifying this file on every release.
On another hand, Git history in master branch gives a picture where the product was released and I think it's not critical but good thing to have. 

@ibuziuk Have you considered putting `nightly` version in VERSION file when you defined the release process for the plugin registry? WDYT?

PR #90, opened 2020-01-31
Add RELEASE.md and release automation script

amisevsk:
### What does this PR do?
Adds `RELEASE.md` to document release process and `make-release.sh` script (adapted from che-plugin-registry repo).

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,20 @@
+# Plugin broker release process
+
+Che plugin broker releases are built automatically from webhooks on the `release` branch and uses the version specified in `VERSION` to tag the built images. To create a new release:
+
+1. Create a `major.minor.x` branch if necessary (e.g. `v3.2.x`)
+2. Update `VERSION` file to refer to the new version to be released, with bugfix version number included (e.g. `v3.2.0`)
+    - Note the convention for the Che plugin broker is to prefix release versions with `v`, i.e. `v3.2.0` instead of `3.2.0`.
+3. Push branch to main repo
+4. Reset the `release` branch to the head of your release branch and push it to Github to trigger CI.
+5. For non-bugfix releases, bump the major/minor version in VERSION on the master branch
+
+For bugfix releases, the `major.minor.x` branch should be reused, with necessary commits cherry-picked into it.
+
+## Automation
+There is a script to automate the release process, named `make-release.sh`:
+```
+./make-release.sh --repo git@github.com:eclipse/che-plugin-registry --version 3.2.1 --trigger-release
+```
+
+See `./make-release.sh --help` for usage.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,6 +1,6 @@
 # Plugin broker release process
 
-Che plugin broker releases are built automatically from webhooks on the `release` branch and uses the version specified in `VERSION` to tag the built images. To create a new release:
+Che plugin broker releases are [built automatically](https://ci.centos.org/job/devtools-che-plugin-broker-release/) from webhooks on the `release` branch and use the version specified in `VERSION` to tag the built images. To create a new release:
 
 1. Create a `major.minor.x` branch if necessary (e.g. `v3.2.x`)
 2. Update `VERSION` file to refer to the new version to be released, with bugfix version number included (e.g. `v3.2.0`)
@@ -14,7 +14,7 @@ For bugfix releases, the `major.minor.x` branch should be reused, with necessary
 ## Automation
 There is a script to automate the release process, named `make-release.sh`:
 ```
-./make-release.sh --repo git@github.com:eclipse/che-plugin-registry --version 3.2.1 --trigger-release
+./make-release.sh --repo git@github.com:eclipse/che-plugin-broker --version 3.2.1 --trigger-release
 ```
 
 See `./make-release.sh --help` for usage.
--- a/make-release.sh
+++ b/make-release.sh
@@ -3,10 +3,23 @@
 # Used to create branch/tag and update VERSION files
 # and and trigger release by force pushing changes to the release branch
 
+RED='\033[0;31m'
+GREEN='\033[0;32m' # TODO
+NC='\033[0m'
+
 # set to 1 to actually trigger changes in the release branch
-TRIGGER_RELEASE=0
-NOCOMMIT=0
-MAJOR_VERSION_BUMP=0
+TRIGGER_RELEASE="false"
+
+# Wrapper for git commands
+function gitw() {
+  if [[ "$1" == "push" ]] && [[ "$TRIGGER_RELEASE" != "true" ]]; then
+    echo -e "    ${RED}'git $@'${NC} (dry run)"
+  else
+    echo -e "    ${GREEN}git $@${NC}"
+    git "$@" 2>&1 | sed 's|^|        |g'
+    return "${PIPESTATUS[0]}"
+  fi
+}
 
 usage ()
 {
@@ -14,8 +27,6 @@ usage ()
   echo "Options:"
   echo "    --help, -h"
   echo "        Show this message."
-  echo "    --no-commit, -n"
-  echo "        Do not commit changes."
   echo "    --trigger-release, -t"
   echo "        Push changes to remote repo, triggering release."
   echo "Example: $0 --repo git@github.com:eclipse/che-subproject --version 'v3.2.1' --trigger-release"; echo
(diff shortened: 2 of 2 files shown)


make-release.sh
```diff
@@ -0,0 +1,159 @@
...
+  git checkout "${BRANCH}"
+  git tag "${VERSION}"
+  git push origin "${VERSION}"
+fi
+
+# now update ${BASEBRANCH} to the new snapshot version
+git fetch origin "${BASEBRANCH}":"${BASEBRANCH}"
+git checkout "${BASEBRANCH}"
+
+# change VERSION file + commit change into ${BASEBRANCH} branch
```
nickboldt: I see you've spiffed up my script to add support for a hypothetical 4.0.0 release. Planning more breaking API changes any time soon? :)

RELEASE.md
```diff
@@ -0,0 +1,20 @@
+# Plugin broker release process
+
+Che plugin broker releases are built automatically from webhooks on the `release` branch and uses the version specified in `VERSION` to tag the built images. To create a new release:
+
+1. Create a `major.minor.x` branch if necessary (e.g. `v3.2.x`)
+2. Update `VERSION` file to refer to the new version to be released, with bugfix version number included (e.g. `v3.2.0`)
+    - Note the convention for the Che plugin broker is to prefix release versions with `v`, i.e. `v3.2.0` instead of `3.2.0`.
```
nickboldt: Because... daring to be different? :P

RELEASE.md
```diff
@@ -0,0 +1,20 @@
+# Plugin broker release process
+
+Che plugin broker releases are built automatically from webhooks on the `release` branch and uses the version specified in `VERSION` to tag the built images. To create a new release:
```
nickboldt: Should we link to the CI job, like in other projects?

RELEASE.md
```diff
@@ -0,0 +1,20 @@
...
+3. Push branch to main repo
+4. Reset the `release` branch to the head of your release branch and push it to Github to trigger CI.
+5. For non-bugfix releases, bump the major/minor version in VERSION on the master branch
+
+For bugfix releases, the `major.minor.x` branch should be reused, with necessary commits cherry-picked into it.
+
+## Automation
+There is a script to automate the release process, named `make-release.sh`:
+```
+./make-release.sh --repo git@github.com:eclipse/che-plugin-registry --version 3.2.1 --trigger-release
```
nickboldt: plugin-registry?

RELEASE.md
nickboldt: but it's cool. some projects use a x.y.z version as a branch. A branch! https://github.com/che-incubator/chectl/tree/7.7.1 << what is that madness?

make-release.sh
amisevsk: Always and forever ;)

The broker changes major version relatively often (at least compared to Che server), so I had a choice between "check if major is incremented and fail if so" or "check if major is implemented and handle it". I did option 1 and realized I may as well support it.

I predict container merging will warrant a v4 since it'll change the structure of the output.

RELEASE.md
amisevsk: We mostly keep doing it because that's how we started doing it. `v3.0.0` was released so it wouldn't make sense to switch conventions for `3.1.0`, etc.

RELEASE.md
amisevsk: :+1: Good idea

RELEASE.md
amisevsk: now that's a copy-paste error for ya

amisevsk:
@nickboldt Added a quick check to see if hub is installed and prompt the user to manually open the last PR if it's not.

RELEASE.md
sleshchenko: Well, I believe leading v is a requirement from Golang.
There are workarounds we could use if case versioning without `v` but they are workarounds.

>  valid semver tags (including the required leading v)
from https://github.com/golang/go/wiki/Modules#can-a-module-consume-a-package-that-has-not-opted-in-to-modules

make-release.sh
```diff
@@ -0,0 +1,164 @@
...
+fi
+
+# change VERSION file
+echo "${VERSION}" > VERSION
+
+# commit change into branch
+if [[ ${NOCOMMIT} -eq 0 ]]; then
+  COMMIT_MSG="[release] Bump to ${VERSION} in ${BRANCH}"
+  git commit -s -m "${COMMIT_MSG}" VERSION
+  git pull origin "${BRANCH}"
```
sleshchenko: If I understand correctly - it's here to update the branch before pushing and it can produce merge commit appearing.
I believe it should be stable enough if we move pulling before committing command.
Then we higher chance to get our git push command failed, but the value of this increment like up to second. And as small profit out history will be liner.
WDYT?

make-release.sh
```diff
@@ -0,0 +1,164 @@
...
+  git push origin "${VERSION}"
+fi
+
+# now update ${BASEBRANCH} to the new snapshot version
+git fetch origin "${BASEBRANCH}":"${BASEBRANCH}"
+git checkout "${BASEBRANCH}"
+
+# change VERSION file + commit change into ${BASEBRANCH} branch
+if [[ ${MAJOR_VERSION_BUMP} -eq 1 ]]; then
+  NEXTVERSION="${NEW_MAJOR_VERSION}.1.0-SNAPSHOT"
```
sleshchenko: `NEXTVERSION` -> `NEXT_VERSION`?

make-release.sh
```diff
@@ -0,0 +1,164 @@
...
+    '-r'|'--repo') REPO="$2"; shift 1;;
+    '-v'|'--version') VERSION="$2"; shift 1;;
+    '-t'|'--trigger-release')
+      if [[ ! ${NOCOMMIT} -eq 0 ]]; then
+        echo "Only one of --trigger-release and --no-commit may be specified"
+        exit 1
+      fi
+      TRIGGER_RELEASE=1; shift 0;;
+    '-n'|'--no-commit')
+      if [[ ! ${NOCOMMIT} -eq 0 ]]; then
```
sleshchenko:
```suggestion
      if [[ ! ${TRIGGER_RELEASE} -eq 0 ]]; then
```
?

make-release.sh
```diff
@@ -0,0 +1,164 @@
...
+  BASEBRANCH="${BRANCH}"
+fi
+
+# work in tmp dir
+TMP=$(mktemp -d); pushd "$TMP" > /dev/null || exit 1
+
+# get sources from ${BASEBRANCH} branch
+echo "Check out ${REPO} to ${TMP}/${REPO##*/}"
+git clone "${REPO}" -q
+cd "${REPO##*/}" || exit 1
```
sleshchenko:
```suggestion
cd ./* || exit 1
```
It should help to solve an issue when user copied/pasted URL from GitHub when ` git` appears in the end, like me
```
❯ ./make-release.sh --repo git@github.com:sleshchenko/che-plugin-broker.git --version v3.3.0 --no-commit
Check out git@github.com:sleshchenko/che-plugin-broker.git to /tmp/tmp.jNL8XGtytZ/che-plugin-broker.git
./make-release.sh: line 77: cd: che-plugin-broker.git: No such file or directory

❯ cd /tmp/tmp.jNL8XGtytZ/che-plugin-broker.git
cd: no such file or directory: /tmp/tmp.jNL8XGtytZ/che-plugin-broker.git

❯ cd /tmp/tmp.jNL8XGtytZ/che-plugin-broker/
# Success!!!!
```

make-release.sh
```diff
@@ -0,0 +1,164 @@
...
+  case $1 in
+    '-r'|'--repo') REPO="$2"; shift 1;;
+    '-v'|'--version') VERSION="$2"; shift 1;;
+    '-t'|'--trigger-release')
+      if [[ ! ${NOCOMMIT} -eq 0 ]]; then
+        echo "Only one of --trigger-release and --no-commit may be specified"
+        exit 1
+      fi
+      TRIGGER_RELEASE=1; shift 0;;
+    '-n'|'--no-commit')
```
sleshchenko: What is the purpose of the no-commit option?
Is it something like dry-run? If yes - let's revise if output if full to understand if the script works correctly, now:
Screenshot_20200203_143834


Also, maybe it would be better to move to dry-run option which would print all operation it going to do with remotes but without performing it, like:
```
  Creating remote branch BRANCH_NAME (v3.2.x) for repo URL
  Bumping version on master to `v3.3.0-SNAPHOT`
  Creating PR ...
```

make-release.sh
```diff
@@ -0,0 +1,164 @@
...
+  BRANCH=${BASEBRANCH}
+  # commit change into branch
+  COMMIT_MSG="[release] Bump to ${NEXTVERSION} in ${BRANCH}"
+  git commit -s -m "${COMMIT_MSG}" VERSION
+  git pull origin "${BRANCH}"
+
+  PUSH_TRY="$(git push origin "${BRANCH}")"
+  # shellcheck disable=SC2181
+  if [[ $? -gt 0 ]] || [[ $PUSH_TRY == *"protected branch hook declined"* ]]; then
+  PR_BRANCH=pr-master-to-${NEXTVERSION}
```
sleshchenko:
```suggestion
    PR_BRANCH=pr-master-to-${NEXTVERSION}
```

make-release.sh
amisevsk: :+1: good idea.

make-release.sh
amisevsk: Good catch, thanks.

make-release.sh
amisevsk: Makes sense to me.

make-release.sh
amisevsk: That would make more sense, this is mostly from copying the script in from other repos. @nickboldt WDYT?

nickboldt: `nocommit` was essentially a `dry-run` option so I could generate plugin registry content w/o pushing commits, in order to verify what was generated before pushing

For that particular repo, we should really have a verification test that runs all the bash scripts from build/script/* in the correct order to make sure that generated yaml is correct.

And in fact since this script has now been copied 4 times to 4 repos, we might want to consider centralizing it into an upstream Che repo. Maybe che-parent? or a new che-release repo? Would be nice to not have the same boilerplate script across 4 repos... as a variation will be needed for che-parent, che-docs, and che repos too in the near future.

make-release.sh
nickboldt: just a nitpick, but why do a if "not equal 0" when you can do `if [[ ${var} -gt 0 ]]` ?

make-release.sh
```diff
@@ -0,0 +1,181 @@
...
+  # create pull request for master branch, as branch is restricted
+  gitw checkout -b "${PR_BRANCH}"
+  gitw push origin "${PR_BRANCH}"
+  lastCommitComment="$(git log -1 --pretty=%B)"
+  if command -v hub &> /dev/null; then
+    hub pull-request -o -f -m "${lastCommitComment}
+
+${lastCommitComment}" -b "${BASE_BRANCH}" -h "${PR_BRANCH}"
+  else
+    echo "Could not open PR in $REPO: 'hub' not installed"
```
nickboldt: We could check that hub is installed BEFORE getting this far. That way if we're triggering this script from Jenkins, we'll fail faster for a node that doesn't have hub installed. Better that than a half-release (branch created, updated, tagged; master branch not changed).

nickboldt: could be something as simple as https://github.com/redhat-developer/codeready-workspaces/blob/master/product/updateBaseImages.sh#L9-L17 but maybe using a `which` check so that install path is more flexible.

amisevsk: I don't want to make `hub` a hard requirement for the script though, since I don't want to install hub locally on my machine.

nickboldt:
if we plan to move this into a central repo for reuse w/ machine exec and devfile registry, and have the option to call additional scripts, then making hub a prereq might be valuable

but for now this is good enough and is merged.

PR #92, opened 2020-02-04
Add CPU limit\request and memory request into container model

mshaposhnik:
### What does this PR do?
Add CPU limit\request and memory request into plugin container model

https://github.com/eclipse/che/issues/13709

### DIFF
--- a/brokers/testdata/theia-7.4.0.yaml
+++ b/brokers/testdata/theia-7.4.0.yaml
@@ -66,7 +66,10 @@ spec:
          - exposedPort: 13131
          - exposedPort: 13132
          - exposedPort: 13133
+     memoryRequest: "256M"
      memoryLimit: "512M"
+     CPURequest: "500m"
+     CPULimit: "1000m"
   initContainers:
   - name: remote-runtime-injector
     image: eclipse/che-theia-endpoint-runtime-binary:7.4.0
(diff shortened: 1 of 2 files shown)


brokers/testdata/theia-7.4.0.yaml
```diff
@@ -66,7 +66,10 @@ spec:
          - exposedPort: 13131
          - exposedPort: 13132
          - exposedPort: 13133
+     memoryRequest: "256M"
      memoryLimit: "512M"
+     CPURequest: "500m"
```
skabashnyuk: cpuRequest
cpuLimit
?

model/model.go
```diff
@@ -92,16 +92,19 @@ type ExposedPort struct {
...
-	Command      []string      `json:"command" yaml:"command"`
-	Args         []string      `json:"args" yaml:"args"`
+	Name          string        `json:"name,omitempty" yaml:"name,omitempty"`
+	Image         string        `json:"image,omitempty" yaml:"image,omitempty"`
+	Env           []EnvVar      `json:"env" yaml:"env"`
+	Commands      []Command     `json:"commands" yaml:"commands"`
+	Volumes       []Volume      `json:"volumes" yaml:"volumes"`
+	Ports         []ExposedPort `json:"ports" yaml:"ports"`
+	CPULimit      string        `json:"CPULimit,omitempty" yaml:"CPULimit,omitempty"`
+	CPURequest    string        `json:"CPURequest,omitempty" yaml:"CPURequest,omitempty"`
```
amisevsk: I would prefer these serialize to `cpuLimit`/`cpuRequest` to match other fields (e.g. `apiVersion` instead of `APIVersion`).

PR #93, opened 2020-02-11
Check plugins' extensions for conflicts during metadata broker

amisevsk:
### What does this PR do?
During metadata brokering process, check plugins for conflicting extensions (i.e. two plugins use the same extension), and warn user in case of conflict. 

Sample output:
```
List of plugins and editors to install
- redhat/java11/latest - Java Linting, Intellisense, formatting, refactoring, Maven/Gradle support and more...
- ms-kubernetes-tools/vscode-kubernetes-tools/latest - Develop, deploy and debug Kubernetes applications
- redhat/vscode-openshift-connector/latest - Interacting with Red Hat OpenShift clusters and providing a streamlined developer experience using Eclipse Che
- eclipse/che-machine-exec-plugin/7.8.0 - Che Plug-in with che-machine-exec service to provide creation terminal or tasks for Eclipse CHE workspace containers.
- eclipse/che-theia/7.8.0 - Eclipse Theia
Warning: some plugins in this workspace have conflicting dependencies:
    Plugins
        {ms-kubernetes-tools/vscode-kubernetes-tools/latest, redhat/vscode-openshift-connector/latest}
    conflict on extension
        https://download.jboss.org/jbosstools/vscode/3rdparty/vscode-kubernetes-tools/vscode-kubernetes-tools-1.0.9.vsix
    Plugins
        {ms-kubernetes-tools/vscode-kubernetes-tools/latest, redhat/vscode-openshift-connector/latest}
    conflict on extension
        https://github.com/redhat-developer/vscode-yaml/releases/download/0.4.0/redhat.vscode-yaml-0.4.0.vsix
If you encounter issues with plugins, please disable conflicting plugins above.
All plugin metadata has been successfully processed
```

The current approach is limited to matching URLs; as a result it can miss some conflicts (e.g. different URLs for the same extension or different versions of the same extension).

Additionally, due to https://github.com/eclipse/che/issues/15758, sometimes the order of messages can sometimes be confused slightly.

Component of https://github.com/eclipse/che/issues/15966

### DIFF
--- a/utils/dependencies.go
+++ b/utils/dependencies.go
@@ -1,0 +1,56 @@
+//
+// Copyright (c) 2020 Red Hat, Inc.
+// This program and the accompanying materials are made
+// available under the terms of the Eclipse Public License 2.0
+// which is available at https://www.eclipse.org/legal/epl-2.0/
+//
+// SPDX-License-Identifier: EPL-2.0
+//
+// Contributors:
+//   Red Hat, Inc. - initial API and implementation
+//
+
+package utils
+
+import (
+	"fmt"
+
+	"github.com/eclipse/che-plugin-broker/model"
+)
+
+// GetExtensionCollisions checks a list of plugin metas for extensions shared by
+// more than one plugin. Return value is a list of
+func GetExtensionCollisions(metas []model.PluginMeta) map[string][]string {
+	extensions := map[string][]string{}
+	for _, meta := range metas {
+		for _, ext := range meta.Spec.Extensions {
+			extensions[ext] = append(extensions[ext], meta.ID)
+		}
+	}
+	collisions := make(map[string][]string)
+	for ext, ids := range extensions {
+		if len(ids) > 1 {
+			collisions[ext] = ids
+		}
+	}
+	return collisions
+}
+
+// ConvertCollisionsToLog converts the output of GetExtensionCollisions to a human-readable
+// string. Output is a slice of strings, to be joined by newlines
+func ConvertCollisionsToLog(collisions map[string][]string) []string {
+	var output []string
+	for ext, plugins := range collisions {
+		output = append(output, fmt.Sprintf("  Plugins"))
+		for _, plugin := range plugins {
+			output = append(output, fmt.Sprintf("    - %s", plugin))
+		}
+		if len(plugins) > 2 {
+			output = append(output, fmt.Sprintf("  all depend on and embed extension"))
+		} else {
+			output = append(output, fmt.Sprintf("  both depend on and embed extension"))
+		}
+		output = append(output, fmt.Sprintf("    %s", ext))
+	}
+	return output
+}
--- a/brokers/metadata/broker.go
+++ b/brokers/metadata/broker.go
@@ -28,7 +28,6 @@ import (
 const RegistryURLFormat = "%s/%s/meta.yaml"
 
 // Broker is used to process Che plugins
-// Broker is used to process Che plugins
 type Broker struct {
 	common.Broker
 	Storage          storage.Storage
@@ -71,6 +70,13 @@ func (b *Broker) Start(pluginFQNs []model.PluginFQN, defaultRegistry string) err
 		return b.fail(fmt.Errorf("Failed to download plugin meta: %s", err))
 	}
 	b.PrintPlan(pluginMetas)
+
+	if collisions := utils.GetExtensionCollisions(pluginMetas); len(collisions) > 0 {
+		collisionLog := []string{"WARNING: multiple instances of the same extension will be included in this workspace:"}
+		collisionLog = append(collisionLog, utils.ConvertCollisionsToLog(collisions)...)
+		collisionLog = append(collisionLog, "These plugins may not work as expected. If errors occur please try disabling all but one of the conflicting plugins.")
+		b.PrintInfoBuffer(collisionLog)
+	}
 
 	// Process plugins into ChePlugins
 	err = b.ProcessPlugins(pluginMetas)
--- a/brokers/testdata/config-plugin-ids.json
+++ b/brokers/testdata/config-plugin-ids.json
@@ -3,6 +3,9 @@
     "id": "ms-kubernetes-tools/vscode-kubernetes-tools/0.1.17"
   },
   {
+    "id": "redhat/vscode-openshift-connector/0.1.2"
+  },
+  {
     "id": "eclipse/che-machine-exec-plugin/0.0.1"
   },
   {
--- a/utils/dependencies_test.go
+++ b/utils/dependencies_test.go
@@ -1,0 +1,86 @@
+//
+// Copyright (c) 2020 Red Hat, Inc.
+// This program and the accompanying materials are made
+// available under the terms of the Eclipse Public License 2.0
+// which is available at https://www.eclipse.org/legal/epl-2.0/
+//
+// SPDX-License-Identifier: EPL-2.0
+//
+// Contributors:
+//   Red Hat, Inc. - initial API and implementation
+//
+
+package utils
+
+import (
+	"regexp"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/eclipse/che-plugin-broker/model"
+)
+
+func TestGetExtensionCollisions(t *testing.T) {
+	metas := []model.PluginMeta{
+		generateMetaWithExtensions("test/conflict_one", "aaa", "bbb"),
+		generateMetaWithExtensions("test/no_conflict", "ddd"),
+		generateMetaWithExtensions("test/conflict_two", "ccc", "bbb"),
+	}
+
+	output := GetExtensionCollisions(metas)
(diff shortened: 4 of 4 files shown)


nickboldt:
Good idea, error message needs work.

```
Warning: some plugins in this workspace have conflicting dependencies:
    Plugins
        {ms-kubernetes-tools/vscode-kubernetes-tools/latest, redhat/vscode-openshift-connector/latest}
    conflict on extension
        https://download.jboss.org/jbosstools/vscode/3rdparty/vscode-kubernetes-tools/vscode-kubernetes-tools-1.0.9.vsix
    Plugins
        {ms-kubernetes-tools/vscode-kubernetes-tools/latest, redhat/vscode-openshift-connector/latest}
    conflict on extension
        https://github.com/redhat-developer/vscode-yaml/releases/download/0.4.0/redhat.vscode-yaml-0.4.0.vsix
```

You should identify WHY it's a conflict... if both k8s and os plugins use the same 1.0.9 version of k8s-tools, then why is it a conflict? I'd suggest something more like:

```
The following plugins require more than one version of the same extension:
* ms-kubernetes-tools/vscode-kubernetes-tools/latest -> vscode-kubernetes-tools-1.0.9.vsix
* redhat/vscode-openshift-connector/latest -> vscode-kubernetes-tools-1.50.59.vsix

* ms-kubernetes-tools/vscode-kubernetes-tools/latest -> redhat.vscode-yaml-0.4.0.vsix
* redhat/vscode-openshift-connector/latest -> redhat.vscode-yaml-0.54.50.vsix
```

Also is your call to action even doable?

```
If you encounter issues with plugins, please disable conflicting plugins above.
All plugin metadata has been successfully processed
```

eg., if you have a devfile which loads a couple plugins (like k8s and yml), how can you reorganize the devfile to load only ONE version of an extension? Do you need to rebuild the plugin registry to not list more than one extension? 

What if we remodel the plugin registry so that it no longer has this 1:1:N mapping of 1 plugin definition to 1 sidecar to N vsix files? If things we 1:1:1 then a user would just enable or disable plugins without having to know about extensions. Or even better... could we add vsix files directly as plugins without the need to wrap them in extra metadata and a sidecar? (I suspect no, but thought it might be worth considering for SOME simple no-runtime-needed extensions like the yaml one.)

utils/dependencies.go
```diff
@@ -0,0 +1,52 @@
...
+	return collisions
+}
+
+// ConvertCollisionsToLog converts the ouput of GetExtensionCollisions to a human-readable
+// string. Output is a slice of strings, to be joined by newlines
+func ConvertCollisionsToLog(collisions map[string][]string) []string {
+	var output []string
+	for ext, plugins := range collisions {
+		output = append(output,
+			fmt.Sprintf("    Plugins"),
```
sleshchenko: Does it make sense to cover it with tests to make sure that there are no typos in the output?
Sometimes I figure out them in messages during test writing.

Also, I wonder what is the better format:
```
  Plugins 
    {ms-kubernetes-tools/vscode-kubernetes-tools/0.1.17, redhat/vscode-openshift-connector/0.1.2} 
  conflict on extension 
    https://github.com/redhat-developer/vscode-yaml/releases/download/0.4.0/redhat.vscode-yaml-0.4.0.vsix
```
vs
```
  Plugins 
    - ms-kubernetes-tools/vscode-kubernetes-tools/0.1.17
    - redhat/vscode-openshift-connector/0.1.2
  conflict on extension 
    https://github.com/redhat-developer/vscode-yaml/releases/download/0.4.0/redhat.vscode-yaml-0.4.0.vsix
```

amisevsk: I'm generally against testing log messages for exact structure, since it doesn't really do much once they're written; any change to the log message means blindly mirroring the message to the test. Either way, I've added a test to ensure all elements are *somewhere* in the log (all conflicting plugins are mentioned).

Also reworked the log message based on your suggestion.

utils/dependencies.go
```diff
@@ -0,0 +1,52 @@
...
+	collisions := make(map[string][]string)
+	for ext, ids := range extensions {
+		if len(ids) > 1 {
+			collisions[ext] = ids
+		}
+	}
+	return collisions
+}
+
+// ConvertCollisionsToLog converts the ouput of GetExtensionCollisions to a human-readable
```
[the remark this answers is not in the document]
amisevsk: :+1: good catch, thanks.

amisevsk:
Updated the log message to be a little cleaner:
```
Warning: some plugins in this workspace have conflicting dependencies:
  Plugins
    - ms-kubernetes-tools/vscode-kubernetes-tools/0.1.17
    - redhat/vscode-openshift-connector/0.1.2
  both depend on extension
    https://github.com/redhat-developer/vscode-yaml/releases/download/0.4.0/redhat.vscode-yaml-0.4.0.vsix
If you encounter issues with plugins, please disable conflicting plugins above.
```

amisevsk:
@nickboldt I think you're imagining my fixup is better than it actually is :)

This PR just checks if two plugins use the *exact* same dependency (*including source URL*). As for suggesting a better remediation than "turn one of these off", there's not much we can do at the moment. At least the warning will give us something to work off of in bug reports. Your example

> ```
> The following plugins require more than one version of the same extension:
> * ms-kubernetes-tools/vscode-kubernetes-tools/latest -> vscode-kubernetes-tools-1.0.9.vsix
> * redhat/vscode-openshift-connector/latest -> vscode-kubernetes-tools-1.50.59.vsix
>
> * ms-kubernetes-tools/vscode-kubernetes-tools/latest -> redhat.vscode-yaml-0.4.0.vsix
> * redhat/vscode-openshift-connector/latest -> redhat.vscode-yaml-0.54.50.vsix
> ```

will not even be noticed by this PR.

Further improvements depend on https://github.com/eclipse/che/issues/15376 and is a part of https://github.com/eclipse/che/issues/15966.

Note also that for most of our plugins, the relationship would be strict subsets (e.g. the extensions in `vscode-yaml` are a subset of the extensions in `kuberentes-tooling` are a subset of the extensions in `openshift-connector`). In that case, enabling `openshift-connector` gives you yaml and kubernetes support. The only exception to this is the camelk extension, which depends on kubernetes-tooling, making it incompatible with openshift-connector. See also the description of PR https://github.com/eclipse/che-plugin-registry/pull/355

amisevsk:
@l0rd @benoitf Before I merge and include this in 7.9.0, I'd like to get your signoff on this approach. Is this sufficient for now, given the state of plugin deps? Should we be doing more to guide the user?

nickboldt:
My only remaining confusion w/ the updated warning message is that it talks about "conflicting dependencies" but the dependencies are to the SAME plugin... so why is that conflicting? 

If the issue is that the user has multiple references to `latest` which could move unexpectedly, sure. But why is 2 refs to the same plugin *at the same URL* a problem?

benoitf:
It's because you can only deploy one instance of plugin. 

Deploying pluginA on sidecar1 and pluginA on sidecar2 is an error ( pluginA being a dependency on some other plugins for example)

nickboldt:
So which sidecar gets the plugin? First come first plugged? Alphabetical? Random hash?

benoitf:
It's random

l0rd:
A few change proposals to the warning text (otherwise LGTM):
- `WARNING` instead of `Warning`
- "multiple instances of the same extension will be included in this workspace" instead of "some plugins in this workspace have conflicting dependencies"
- "both depend and embed" instead of "both depend"
- "As a consequence those plugins may not work as expected. If that's the case try disabling all but one of the conflicting plugins" instead of "If you encounter issues with plugins, please disable conflicting plugins above."

cc @beaumorley for the message wording

beaumorley:
Is it possible to paste the final message here?

amisevsk:
@beaumorley With Mario's suggested changes, an example output for the broker, shown during workspace startup, would be
```
List of plugins and editors to install
- ms-kubernetes-tools/vscode-kubernetes-tools/0.1.17 - Develop, deploy and debug Kubernetes applications
- redhat/vscode-openshift-connector/0.1.2 - Interacting with Red Hat OpenShift clusters and providing a streamlined developer experience using Eclipse Che
- eclipse/che-machine-exec-plugin/0.0.1 - Che Plug-in with che-machine-exec service to provide creation terminal or tasks for Eclipse CHE workspace machines.
- redhat/java/0.43.0 - Java Linting, debugging, Intellisense, formatting, refactoring, Maven/Gradle support and more...
- eclipse/che-theia/next - Eclipse Theia, get the latest release each day.
WARNING: multiple instances of the same extension will be included in this workspace:
  Plugins
    - ms-kubernetes-tools/vscode-kubernetes-tools/0.1.17
    - redhat/vscode-openshift-connector/0.1.2
  both depend on and embed extension
    https://github.com/redhat-developer/vscode-yaml/releases/download/0.4.0/redhat.vscode-yaml-0.4.0.vsix
These plugins may not work as expected. If errors occur please try disabling all but one of the conflicting plugins.
All plugin metadata has been successfully processed
```

beaumorley:
thanks @amisevsk lgtm!

PR #94, opened 2020-02-11
Cleanup

amisevsk:
### What does this PR do?
Cleanup brokers codebase:
- Set license years to 2020
- Use quay.io image registry in makefile to match where images are normally pushed.

### DIFF
--- a/.cico/cico_functions.sh
+++ b/.cico/cico_functions.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# Copyright (c) 2012-2019 Red Hat, Inc.
+# Copyright (c) 2019-2020 Red Hat, Inc.
 # This program and the accompanying materials are made
 # available under the terms of the Eclipse Public License 2.0
 # which is available at https://www.eclipse.org/legal/epl-2.0/
--- a/Makefile
+++ b/Makefile
@@ -39,11 +39,11 @@ dep-update:
 
 .PHONY: build-docker-artifacts
 build-docker-artifacts:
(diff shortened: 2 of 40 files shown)


.cico/cico_functions.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# Copyright (c) 2012-2019 Red Hat, Inc.
+# Copyright (c) 2012-2020 Red Hat, Inc.
```
nickboldt: has this really existed since 2012? I've been using 2018-2020 for most things...

Makefile
```diff
@@ -39,11 +39,11 @@ dep-update:
 
 .PHONY: build-docker-artifacts
 build-docker-artifacts:
-	docker build -t eclipse/che-plugin-artifacts-broker:latest -f build/artifacts/Dockerfile .
+	docker build -t quay.io/eclipse/che-plugin-artifacts-broker:latest -f build/artifacts/Dockerfile .
 
 .PHONY: build-docker-metadata
 build-docker-metadata:
-	docker build -t eclipse/che-plugin-metadata-broker:latest -f build/metadata/Dockerfile .
+	docker build -t quay.io/eclipse/che-plugin-metadata-broker:latest -f build/metadata/Dockerfile .
```
nickboldt: +1, this is good.

.cico/cico_functions.sh
amisevsk: Huh, I just ran a find-replace. I wonder if we can track down the original `cico_functions.sh` -- it generally just gets copied in for everything that runs in centos CI.

PR #98, opened 2020-03-17
Lifecycle

vparfonov:
### What does this PR do?
Add Lifecycle to the model. This should allow `postStart` and `preStop` actions for plugin containers. More information here: https://kubernetes.io/docs/tasks/configure-pod-container/attach-handler-lifecycle-event/ 

In this PR will be introduced only [`ExecAction`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#execaction-v1-core). 
This feature looks useful for some operation that must be called just after start or before stop container. For example some will be used for backup (as `preStop` action) and restore (as `postStart`) project source.

[HTTPGetAction](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#httpgetaction-v1-core) and [TCPSocketAction](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#tcpsocketaction-v1-core) not implement yet.

Docker image for testing:
`docker pull vparfonov/che-plugin-metadata-broker:v3.1.2`

### DIFF
--- a/model/model.go
+++ b/model/model.go
@@ -96,7 +96,7 @@ type ExecAction struct {
 }
 
 type Handler struct {
-	Exec  *ExecAction `json:"exec,omitempty" yaml:"exec,omitempty"`
+	Exec *ExecAction `json:"exec,omitempty" yaml:"exec,omitempty"`
 }
 
 type Lifecycle struct {
@@ -118,7 +118,7 @@ type Container struct {
 	MountSources  bool          `json:"mountSources" yaml:"mountSources"`
 	Command       []string      `json:"command" yaml:"command"`
 	Args          []string      `json:"args" yaml:"args"`
-    Lifecycle     *Lifecycle     `json:"lifecycle,omitempty" yaml:"lifecycle,omitempty"`
+	Lifecycle     *Lifecycle    `json:"lifecycle,omitempty" yaml:"lifecycle,omitempty"`
 }
 
 type ChePlugin struct {
--- a/model/model.go
+++ b/model/model.go
@@ -91,6 +91,19 @@ type ExposedPort struct {
 	ExposedPort int `json:"exposedPort" yaml:"exposedPort"`
 }
 
+type ExecAction struct {
+	Command []string `json:"command,omitempty" yaml:"command,omitempty"`
+}
+
+type Handler struct {
+	Exec  ExecAction `json:"exec,omitempty" yaml:"exec,omitempty"`
+}
+
+type Lifecycle struct {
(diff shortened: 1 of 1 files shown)


ibuziuk:
@vparfonov could you please add some context to the PR description, currently it is not clear why do we need `preStop` / `postStart` events on the plugin-broken end

vparfonov:
> @vparfonov could you please add some context to the PR description, currently it is not clear why do we need `preStop` / `postStart` events on the plugin-broken end

Updated description

model/model.go
```diff
@@ -105,6 +118,7 @@ type Container struct {
 	MountSources  bool          `json:"mountSources" yaml:"mountSources"`
 	Command       []string      `json:"command" yaml:"command"`
 	Args          []string      `json:"args" yaml:"args"`
+    Lifecycle     *Lifecycle     `json:"lifecycle,omitempty" yaml:"lifecycle,omitempty"`
```
sleshchenko, line 121: please execute `go fmt -x ./...`

vparfonov: done

model/model.go
```diff
@@ -91,6 +91,19 @@ type ExposedPort struct {
 	ExposedPort int `json:"exposedPort" yaml:"exposedPort"`
 }
 
+type ExecAction struct {
+	Command []string `json:"command,omitempty" yaml:"command,omitempty"`
```
sleshchenko, line 95: I wonder why it's the only array with `omitempty` specified. Probably it's unneeded here.
So, would be cool if you could test without it.

model/model.go
```diff
@@ -105,6 +118,7 @@ type Container struct {
 	MountSources  bool          `json:"mountSources" yaml:"mountSources"`
 	Command       []string      `json:"command" yaml:"command"`
 	Args          []string      `json:"args" yaml:"args"`
+	Lifecycle     *Lifecycle    `json:"lifecycle,omitempty" yaml:"lifecycle,omitempty"`
```
sleshchenko, line 121: Please consider updating model description in plugin-registry 
https://github.com/eclipse/che-plugin-registry#plugin-meta-yaml-structure
https://github.com/eclipse/che-plugin-registry/blob/master/build/scripts/meta.yaml.schema

model/model.go
vparfonov: I will check

vparfonov: Checked, you are right `omitempty` not needed going to remove it

model/model.go
vparfonov: https://github.com/eclipse/che-plugin-registry/pull/442

PR #107, opened 2020-06-05
Remove rhopp from codeowners.

rhopp:
### What does this PR do?
Removes rhopp from codeowners.


rhopp:
Don't worry. I'll be watching you!
https://gph.is/2x3fr4v

PR #109, opened 2020-08-13
Add command line option for CA bundle directory

mmorhun:
Signed-off-by: Mykola Morhun <mmorhun@redhat.com>

### What does this PR do?
Adds and implements new option `-cadir` which is a path to a dir which contains CA certs bundles. In result plugin brokers trust all the certificates in all the files in the provided directory if any.

https://github.com/eclipse/che/issues/17552

### DIFF
--- a/cfg/cfg.go
+++ b/cfg/cfg.go
@@ -67,6 +67,9 @@ var (
 
 	// SelfSignedCertificateFilePath path to certificate file that should be used while connection establishing
 	SelfSignedCertificateFilePath string
+
+	// CABundleDirPath Path to directory with trusted CA certificates
+	CABundleDirPath string
 )
 
 func init() {
@@ -139,6 +142,12 @@ func init() {
 		"",
 		"Path to Certificate that should be used while connection establishing",
 	)
+	flag.StringVar(
+		&CABundleDirPath,
+		"cadir",
+		"",
+		"Path to directory with trusted CA certificates",
+	)
 }
 
 // Parse parses configuration.
@@ -187,6 +196,9 @@ func Print() {
 	log.Printf("    OwnerId: %s", RuntimeID.OwnerId)
 	if SelfSignedCertificateFilePath != "" {
 		log.Printf("  Self signed certificate %s", SelfSignedCertificateFilePath)
+	}
+	if CABundleDirPath != "" {
+		log.Printf("  CA bundle certificates path %s", CABundleDirPath)
 	}
 }
 
--- a/cfg/cfg.go
+++ b/cfg/cfg.go
@@ -65,10 +65,12 @@ var (
 	// Used as a default registry if a plugin fully-qualified name does not specify a registry.
 	RegistryAddress string
 
-	// SelfSignedCertificateFilePath path to certificate file that should be used while connection establishing to Che server
+	// SelfSignedCertificateFilePath path to certificate file that should be used while connection establishing to Che server.
+	// Usually it contains Che server self-signed certificate.
 	SelfSignedCertificateFilePath string
 
-	// CABundleDirPath Path to directory with trusted CA certificates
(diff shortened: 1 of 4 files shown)


common/connect.go
```diff
@@ -36,22 +38,59 @@ func (tb *tunnelBroadcaster) Accept(e event.E) {
...
+					} else {
+						// Append certs from the file to the system pool
+						if ok := rootCAs.AppendCertsFromPEM(certs); !ok {
+							// Log error and continue
+							log.Printf("Failed to append %q to RootCAs.", file)
+						}
+					}
+				}
+			}
+		}
```
amisevsk, line 93: I feel like this could be implemented more cleanly using `filepath.Walk`.

mmorhun: I had considered using of `filepath.Walk` however decided to go with `ioutil.ReadDir` as we do not need recursive walk here. As about other checks, using `filepath.Walk` doesn't change anything in comparison with `ioutil.ReadDir`.

cfg/cfg.go
```diff
@@ -188,6 +197,9 @@ func Print() {
 	if SelfSignedCertificateFilePath != "" {
 		log.Printf("  Self signed certificate %s", SelfSignedCertificateFilePath)
 	}
+	if CABundleDirPath != "" {
```
sleshchenko, line 200: Having `SelfSignedCertificateFilePath` and `CABundleDirPath` at the same time is a bit confusing... Could we just go with one - CABundleDir?

They have slightly different documentation but in fact the only difference according to
```
// ConfigureCertPool trusts given certificates
// CADirPath is a path to directory with CA certificates files
func ConfigureCertPool(CAFilePath string, CADirPath string) {
```
their type - file and dir... And it's possible to specify both.

mmorhun: I am not sure we should. CAFile is the file which contains Che server self-signed certificate, where CABundle is a set of all trusted CA certificates inside Openshift cluster (however it doesn't has the cert from CAFile - it is not clear to me why, but that's how Openshift works). I agree that the comments might be confusing, thought. I'll update them.

common/connect.go
amisevsk: Yeah, looking at it more it looks like Walk wouldn't be much simpler. I think my issue is actually more the depth of nesting here -- we could simplify this a bit:

```go
for _, file := range files {
	if !file.IsDir() {
		continue
	}
	fileExt := filepath.Ext(file.Name())
	// Look up only for certificate files
	if fileExt == ".crt" || fileExt == ".pem" {
		certsFilePath := filepath.Join(CADirPath, file.Name())
		certs, err := ioutil.ReadFile(certsFilePath)
		if err != nil {
			// Log error and continue
			log.Printf("Failed to read certificate from file %q. Error: %v", certsFilePath, err)
			continue
		} 
		// Append certs from the file to the system pool
		if ok := rootCAs.AppendCertsFromPEM(certs); !ok {
			// Log error and continue
			log.Printf("Failed to append %q to RootCAs.", file)
		}
	}
}
```

mmorhun: Updated this part of code.

common/connect.go
```diff
@@ -36,22 +38,63 @@ func (tb *tunnelBroadcaster) Accept(e event.E) {
 	}
 }
 
-func ConfigureCertPool(customCertificateFilePath string) {
-	// Get the SystemCertPool, continue with an empty pool on error
+// ConfigureCertPool trusts given certificates
+// CAFilePath is a path to file which contains CA certificates.
+//   Usually it contains Che server self-signed certificate.
```
sleshchenko, line 43: Please consider propagating it to a more visible place, like flag description.

mmorhun: ok

AndrienkoAleksandr (approved):

PR #118, opened 2021-06-29
Merge Dockerfile & rhel.Dockerfiles into one

mkuznyetsov:
Signed-off-by: Mykhailo Kuznietsov <mkuznets@redhat.com>

### What does this PR do?
Merge content from both Dockerfiles (for easier maintenance)


sleshchenko (commented):
I haven't tested but changes LGTM

sleshchenko (approved):

build/artifacts/Dockerfile
```diff
@@ -10,17 +10,18 @@
...
 WORKDIR /build/che-plugin-broker/brokers/artifacts/cmd/
 COPY . /build/che-plugin-broker/
-RUN CGO_ENABLED=0 GOOS=linux go build -mod vendor -a -ldflags '-w -s' -installsuffix cgo -o artifacts-broker main.go
+RUN adduser appuser && \
+    CGO_ENABLED=0 GOOS=linux go build -mod vendor -a -ldflags '-w -s' -installsuffix cgo -o artifacts-broker main.go
 
 
-FROM alpine:3.12
+# https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/ubi8-minimal
+FROM registry.access.redhat.com/ubi8-minimal:8.4-200.1622548483
```
benoitf, line 23: Can we use micro image there ?

sleshchenko: I would vote for minimal if it's small enough.
If we go with micro, I assume we'll lose the ability to exec inside a container and debug some issues, but I may be wrong.
Particularly for plugin broker - I haven't debugged it for a while, so my comment may be not applicable here.

benoitf: mcro has bash

benoitf: and the image alpine had sh so I would say we don't change anything

sleshchenko: I had must try micro before sharing such a concern )

amisevsk (approved):

build/artifacts/Dockerfile
nickboldt: switched to micro. let's see what breaks! :D

nickboldt (approved):
it's strange that I can commit changes to someone's PR, then approve it. Feels like a loophole :)

nickboldt:
cherry picked to v3.4.x branch, now trying to figure out how to use this image in Brew...

build/artifacts/Dockerfile
nickboldt: FYI, using the ubi8-micro doesn't work in brew for two reasons:

a) the internally available container name is ubi8/ubi-micro, so need to use that if we want to find the base image
b) some internal process needs to run /bin/rpm on the built container, which fails as ubi8-micro has no package manager installed.

benoitf: well for scratch images it can't run RPM as well so this shouldn't be an issue for micro

nickboldt: I suspect there are different rules for Brew builds when the base image is a scratch, than when they're anything else.

